PR #1, opened 2021-05-18
Extract diagnostics from MCMCChains

devmotion:
I extracted the following files and their history from [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl) with
```
git filter-repo --path src/discretediag.jl --path src/ess.jl --path src/gelmandiag.jl --path src/gewekediag.jl --path src/heideldiag.jl --path src/mcse.jl --path src/rafterydiag.jl --path src/rstar.jl --path --path src/utils.jl --path test/diagnostic_tests.jl --path test/ess_tests.jl --path test/rstar_tests.jl
```

There are still some occurrences  `Chains`, so I'll have to add some additional fixes.


sethaxen:
It's a large PR, so I've not gone through everything, but what I've seen looks reasonable. I like that the git history has been preserved. There are a handful of coverage issues, but those can be fixed in a future PR. On my machine the load time is ~0.86s seconds, just over half of MCMCChains' loading time. Not as lightweight as I thought it might be, but also not heavy.

devmotion:
> There are a handful of coverage issues, but those can be fixed in a future PR.

I fixed multiple bugs and type stability issues in the last commits. I also added some tests for each method but so far only for ESS and Rstar correctness of the implementation is checked (somewhat), all other tests only ensure that the methods can be called and are type stable. So code coverage is misleading and in fact most methods are not tested properly. However, since it is still strictly better than the current state in MCMCChains I don't think it matters for this PR and an initial release.

> On my machine the load time is ~0.86s seconds, just over half of MCMCChains' loading time. Not as lightweight as I thought it might be, but also not heavy.

I didn't expect any major load time improvements here (I think I mentioned this in a previous comment) and it's not the motivation for this package either (at least not for me). We already removed all major dependencies (such as e.g. DataFrames) from MCMCChains, and most remaining dependencies were related to the inference diagnostics and not the `Chains` structure. Dependency-wise probably the main difference is that this package does not depend on AxisArrays.

sethaxen:
Yes I don't see anything that should block this PR or a release.

PR #3, opened 2021-06-13
Format code with JuliaFormatter

devmotion:
This PR ensures that the code actually follows BlueStyle, as indicated in the README. I just ran JuliaFormatter and added a Github action that checks that the format is correct.


sethaxen (approved):
LGTM!

PR #6, opened 2021-06-13
Add documentation

devmotion:
The PR adds some initial documentation and fixes docstrings of `ess_rhat` and `rstar`.


devmotion:
Documentation is built correctly but can't be deployed (so no preview :cry:) since no versions exist.

devmotion:
Let's see if it is just a problem with the preview functionality.

PR #7, opened 2021-06-14
Return actual R* distribution

devmotion:
This PR implements the suggestion in https://github.com/TuringLang/MCMCChains.jl/pull/263: the analytical distribution can be calculated explicitly and no sampling has to be performed.

This PR uses the more general location scale distribution in Distributions 0.25 instead of reimplementing a scaled Poisson-binomial distribution. Additionally, the pdf of the Poisson-binomial distribution is computed lazily in Distributions 0.25 which implies that the pdf values are never computed if one only evaluates the mean or variance of the distribution.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 [compat]
 AbstractFFTs = "0.5, 1"
 DataAPI = "1.6"
-Distributions = "0.24, 0.25"
+Distributions = "0.25"
 MLJModelInterface = "1"
 SpecialFunctions = "0.8, 0.9, 0.10, 1"
 StatsBase = "0.33"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 InferenceDiagnostics = "be115224-59cd-429b-ad48-344e309966f0"
+MLJBase = "a7f614a8-145f-11e9-1d2a-a57a1082229d"
 MLJModels = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
 MLJXGBoostInterface = "54119dfa-1dab-4055-a167-80440f4f7a91"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
@@ -8,6 +9,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 Documenter = "0.27"
 InferenceDiagnostics = "0.1"
+MLJBase = "0.18"
 MLJModels = "0.14"
 MLJXGBoostInterface = "0.1"
 julia = "1.3"
--- a/src/rstar.jl
+++ b/src/rstar.jl
@@ -72,10 +72,7 @@ function rstar(
     # train classifier on training data
     ycategorical = MLJModelInterface.categorical(y)
     fitresult, _ = MLJModelInterface.fit(
-        classifier,
-        verbosity,
-        Tables.table(x[train_ids, :]),
-        ycategorical[train_ids],
+        classifier, verbosity, Tables.table(x[train_ids, :]), ycategorical[train_ids]
     )
 
     # compute predictions on test data
@@ -116,7 +113,7 @@ function rstar_distribution(
 
     # scale distribution to support in `[0, nclasses]`
     nclasses = length(MLJModelInterface.classes(ytest))
-    scaled_distribution = (nclasses // length(predictions)) * distribution
+    scaled_distribution = (nclasses//length(predictions)) * distribution
 
     return scaled_distribution
 end
--- a/src/rstar.jl
+++ b/src/rstar.jl
@@ -18,10 +18,10 @@ The `classifier` has to be a supervised classifier of the MLJ framework (see the
 for a list of supported models). It is trained with a `subset` of the samples. The training
 of the classifier can be inspected by adjusting the `verbosity` level.
 
-If the classifier is probabilistic, i.e., if it outputs probabilities of classes, the scaled
-Poisson-binomial distribution of the ``R^*`` statistic is returned (algorithm 2). If the
-classifier is deterministic, i.e., if it predicts a class, the value of the ``R^*``
-statistic is returned (algorithm 1).
+If the classifier is deterministic, i.e., if it predicts a class, the value of the ``R^*``
+statistic is returned (algorithm 1). If the classifier is probabilistic, i.e., if it outputs
+probabilities of classes, the scaled Poisson-binomial distribution of the ``R^*`` statistic
+is returned (algorithm 2).
 
 !!! note
     The correctness of the statistic depends on the convergence of the `classifier` used
@@ -30,17 +30,35 @@ statistic is returned (algorithm 1).
 # Examples
 
 ```jldoctest rstar
-julia> using MLJModels, Statistics
+julia> using MLJBase, MLJModels, Statistics
 
 julia> XGBoost = @load XGBoostClassifier verbosity=0;
 
 julia> samples = fill(4.0, 300, 2);
 
 julia> chain_indices = repeat(1:3; outer=100);
+```
 
+One can compute the distribution of the ``R^*`` statistic (algorithm 2) with the
+probabilistic classifier.
+
+```jldoctest rstar
 julia> distribution = rstar(XGBoost(), samples, chain_indices);
 
 julia> isapprox(mean(distribution), 1; atol=0.1)
+true
+```
+
+For deterministic classifiers, a single ``R^*`` statistic (algorithm 1) is returned.
+Deterministic classifiers can also be derived from probabilistic classifiers by e.g.
+predicting the mode. In MLJ this corresponds to a pipeline of models.
+
+```jldoctest rstar
+julia> @pipeline XGBoost name=XGBoostDeterministic operation=predict_mode;
+
+julia> value = rstar(XGBoostDeterministic(), samples, chain_indices);
+
+julia> isapprox(value, 1; atol=0.1)
 true
 ```
 
--- a/src/rstar.jl
+++ b/src/rstar.jl
@@ -54,7 +54,7 @@ Deterministic classifiers can also be derived from probabilistic classifiers by 
 predicting the mode. In MLJ this corresponds to a pipeline of models.
 
 ```jldoctest rstar
-julia> @pipeline XGBoost name=XGBoostDeterministic operation=predict_mode;
+julia> @pipeline XGBoost name = XGBoostDeterministic operation = predict_mode;
 
 julia> value = rstar(XGBoostDeterministic(), samples, chain_indices);
 
--- a/src/rstar.jl
+++ b/src/rstar.jl
@@ -8,8 +8,8 @@
         verbosity::Int=0,
     )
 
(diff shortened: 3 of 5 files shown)


devmotion:
I still wonder (https://github.com/TuringLang/MCMCChains.jl/pull/263#issuecomment-784425038)

> if it would make sense to scale the statistic from the interval [0, nchains] to [0, 1] - it seems this would simplify the comparison of estimates for chains with different number of chains and possibly increase the interpretability?

But probably that should not be addressed in this PR (if it is reasonable at all).

trappmartin:
Cool!

cc: @avehtari

devmotion:
Any additional comments or suggestions @yebai @sethaxen?

sethaxen (commented):
This is cool! A few general comments.

I think the biggest advantages of having the whole Rstar distribution are 1) exactly computing the mean instead of approximating it from samples 2) ability to visualize its PMF instead of relying on random samples. Unfortunately, the latter does not work:

```julia
julia> plot(distribution)
ERROR: ArgumentError: invalid index: 1//1 of type Rational{Int64}
```

I thought perhaps the issue was the irrational (and maybe it was), but replacing it with floats didn't help.
```julia
julia> distribution2 = Distributions.LocationScale(float(distribution.μ), float(distribution.σ), distribution.ρ);

julia> plot(distribution2)
ERROR: ArgumentError: invalid index: 1.0 of type Float64
```
Can this be fixed with something in this PR, or must it be done elsewhere?

Second, regarding naming, a user might expect `rstar_distribution` to return a distribution, while `rstar` would return a the Rstar statistic, especially because the paper reserves the use of the term Rstar for the statistic itself. I wonder if it makes sense to stick closer to the paper, making `rstar` return a single Rstar statistic a la Algorithm 1 in the paper and `rstar_distribution` return a distribution instead of samples a la Algorithm 2.

devmotion:
> Can this be fixed with something in this PR, or must it be done elsewhere?

This can't be fixed in this PR but must be addressed in Distributions (it should be sufficient to replace `k + 1` with `Int(k + 1)` in https://github.com/JuliaStats/Distributions.jl/blob/dd75dcf1d771c118875aef33538a4b77e5fef878/src/univariate/discrete/poissonbinomial.jl#L126).

> Second, regarding naming, a user might expect rstar_distribution to return a distribution, while rstar would return a the Rstar statistic, especially because the paper reserves the use of the term Rstar for the statistic itself. I wonder if it makes sense to stick closer to the paper, making rstar return a single Rstar statistic a la Algorithm 1 in the paper and rstar_distribution return a distribution instead of samples a la Algorithm 2.

`rstar_distribution` is an internal function that is not exported. I just chose a somewhat meaningful name. Since the pmf of the Poisson-binomial distribution is only evaluated upon request, in general it is cheaper to return the distribution than to perform sampling as done in the current implementation. So I don't think there's a good reason to keep the current implementation.

devmotion:
More generally, IMO the current approach that unifies algorithm 1 and 2 leads to a cleaner API: algorithm 1 applies to deterministic classifiers whereas algorithm 2 is used for probabilistic ones. In the paper in the second case the resulting distribution of the statistic is estimated by sampling but we can just return the analytical distribution in both cases. Alternatively, one could return not a Dirac distribution but its support for deterministic classifiers - maybe that would be more intuitive. On the other hand, then one can't call `mean` etc. in both cases but only when working with probabilistic classifiers.

devmotion:
@sethaxen I made a PR to Distributions that fixes it (https://github.com/JuliaStats/Distributions.jl/pull/1345). I checked that `plot(distribution)` does not error anymore. However, the resulting plot is incorrect since StatsPlots incorrectly assumes that discrete distributions are integer-valued (https://github.com/JuliaPlots/StatsPlots.jl/blob/6e65c64fb99ca5148989766092a5432391ab1e3b/src/distributions.jl#L25). This is not correct e.g. for discrete location-scale distributions, Dirac distributions, and `DiscreteNonParametric`.

avehtari:
Thanks @trappmartin for tagging me. This is very cool. I had missed the February PR TuringLang/MCMCChains.jl#263
I'm tagging also @ben18785

sethaxen:
> More generally, IMO the current approach that unifies algorithm 1 and 2 leads to a cleaner API: algorithm 1 applies to deterministic classifiers whereas algorithm 2 is used for probabilistic ones. In the paper in the second case the resulting distribution of the statistic is estimated by sampling but we can just return the analytical distribution in both cases. Alternatively, one could return not a Dirac distribution but its support for deterministic classifiers - maybe that would be more intuitive. On the other hand, then one can't call `mean` etc. in both cases but only when working with probabilistic classifiers.
> 
> Edit: Actually `mean` works for real numbers as well but other functions of the Distributions interface won't work.

The final paragraph of Section 3.1.3 explains that the probabilistic classifiers used in Algorithm 1 are made deterministic by predicting the class with the highest probability. No sampling occurs there. That section also compares the two Rstar implementations. They're not exactly equivalent. I still then think it would be good to have both algorithms represented: Algorithm 1 to compute a single Rstar value (I still think it makes sense to call this `rstar` as the authors do) and Algorithm 2 but without sampling to construct the analytic distribution (and to me it makes sense to call this `rstar_distribution`).

> @sethaxen I made a PR to Distributions that fixes it ([JuliaStats/Distributions.jl#1345](https://github.com/JuliaStats/Distributions.jl/pull/1345)). I checked that `plot(distribution)` does not error anymore. However, the resulting plot is incorrect since StatsPlots incorrectly assumes that discrete distributions are integer-valued (https://github.com/JuliaPlots/StatsPlots.jl/blob/6e65c64fb99ca5148989766092a5432391ab1e3b/src/distributions.jl#L25). This is not correct e.g. for discrete location-scale distributions, Dirac distributions, and `DiscreteNonParametric`.

Great! Thanks! I'll look into how to fix the StatsPlots bug later.

trappmartin:
> Great! Thanks! I'll look into how to fix the StatsPlots bug later.
TBH, I think this is a very strange handling of discrete distributions in general. I think it make more sense to use the support (or a subset of it) of the distribution instead of defining some unit range that spans the support.

sethaxen:
> > Great! Thanks! I'll look into how to fix the StatsPlots bug later.
> 
> TBH, I think this is a very strange handling of discrete distributions in general. I think it make more sense to use the support (or a subset of it) of the distribution instead of defining some unit range that spans the support.

Yes, agreed. I'm guessing the plotting code for discrete distributions incorrectly assumes their support will always be the integers.

devmotion:
> The final paragraph of Section 3.1.3 explains that the probabilistic classifiers used in Algorithm 1 are made deterministic by predicting the class with the highest probability. No sampling occurs there. That section also compares the two Rstar implementations. They're not exactly equivalent. I still then think it would be good to have both algorithms represented: Algorithm 1 to compute a single Rstar value (I still think it makes sense to call this rstar as the authors do) and Algorithm 2 but without sampling to construct the analytic distribution (and to me it makes sense to call this rstar_distribution).

I disagree with this suggestion because
- the use of two functions will not increase functionality but only complexity of the API: one can already use algorithm 1 by providing a deterministic classifier and algorithm 2 by providing a probabilistic one
- only the choice of the algorithm depending on the type of the classifier handles probabilistic classifier in a clear user-controlled way: you don't always want a balanced decision boundary (e.g., 0.5 in binary classification) but probably that's what we would have to choose for probabilistic classifiers (using `predict_mode` in MLJ); so it seems cleaner if the user just provides a deterministic version of the probabilistic classifier with a custom decision boundary if she wants to evaluate algorithm 1 with a probabilistic classifier
- in my opinion the name of the function is sufficiently consistent with the paper, in particular if it would return a single value instead of Dirac distribution for deterministic classifiers which I think I would prefer

> Yes, agreed. I'm guessing the plotting code for discrete distributions incorrectly assumes their support will always be the integers.

Yes, and actually Distributions does the same incorrectly. I'm working on some PRs that would fix LocationScale, DiscreteNonParametric, and Dirac.

sethaxen:
> * the use of two functions will not increase functionality but only complexity of the API: one can already use algorithm 1 by providing a deterministic classifier and algorithm 2 by providing a probabilistic one
> 
> * only the choice of the algorithm depending on the type of the classifier handles probabilistic classifier in a clear user-controlled way: you don't always want a balanced decision boundary (e.g., 0.5 in binary classification) but probably that's what we would have to choose for probabilistic classifiers (using `predict_mode` in MLJ); so it seems cleaner if the user just provides a deterministic version of the probabilistic classifier with a custom decision boundary if she wants to evaluate algorithm 1 with a probabilistic classifier

I haven't used MLJ much, so perhaps this is the source of our disagreement. If a user wanted to make `XGBoost` a deterministic classifier predicting the maximum probability class as in the paper using MLJ, is this straightforward to do? And if so, perhaps we should also include that as an example e.g. in the docstring.

devmotion:
I'm not familiar with MLJ but I am convinced that conceptually this is the better design since it gives users much more flexibility and control over how the deterministic predictions are obtained from the classifier.

The following should work if one wants to use `predict_mode` (of course, in this case it would be easier to perform evaluation with `predict_mode` but in contrast to a hard-coded `predict_mode` this can be adapted by the user):
```julia
import MLJBase

struct MyModel{M<:MLJBase.Probabilistic} <: MLJBase.Deterministic
    model::M
end

# `fit` just fits the wrapped model
function MLJBase.fit(model::MyModel, verbosity, X, y)
    return MLJBase.fit(model.model, verbosity, X, y)
end

# predict uses `predict_mode`
function MLJBase.predict(model::MyModel, fitresult, Xnew)
    return MLJBase.predict_mode(model.model, fitresult, Xnew)
end
```
(untested and just copied from the simple example in the documentation: https://alan-turing-institute.github.io/MLJ.jl/dev/simple_user_defined_models/#A-simple-deterministic-regressor)

I would assume that there is a simpler way to achieve the same with a pipeline (https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/) but as I said I am not familiar with MLJ.

sethaxen:
> I'm not familiar with MLJ but I am convinced that conceptually this is the better design since it gives users much more flexibility and control over how the deterministic predictions are obtained from the classifier.

I agree that this is better in that the user can do more with it. My concern is just that the average user will probably want to do something similar to the algorithms in the paper, and the added flexibility should not make this cumbersome. A user should not need to define a custom `MLJ` model to replicate the paper's Rstar.

> I would assume that there is a simpler way to achieve the same with a pipeline (https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/) but as I said I am not familiar with MLJ.

Looks like one can do `XGBoost_deterministic = @pipeline(XGBoost, operation=predict_mode)` and then pass `XGBoost_deterministic` to `rstar` in place of `XGBoost()`. I do think this should be a documented example so that the user searching for something equivalent to what the paper did in Algorithm 1 knows how to do it, and with that I would be happy with this API. Do you agree?

devmotion:
> Looks like one can do `XGBoost_deterministic = @pipeline(XGBoost, operation=predict_mode)` and then pass `XGBoost_deterministic` to `rstar` in place of `XGBoost()`. I do think this should be a documented example so that the user searching for something equivalent to what the paper did in Algorithm 1 knows how to do it, and with that I would be happy with this API. Do you agree?

Ah yeah now I remember that I read something about `operation` in the docs :smile: This is a good suggestion, I added it to the tests and explained it in the docstring (I used a slightly different syntax `@pipeline XGBoost name = XGBoostDeterministic operation = predict_mode` that creates a pipeline of type `XGBoostDeterministic`).

sethaxen (approved):
LGTM!

PR #12, opened 2021-06-21
Speed ups to ess_rhat

sethaxen:
This PR adds some speed-ups to `mean_autocov`, which in turn speeds up `ess_rhat` for two ess methods.

### `ESSMethod`:

`eltype(samples) isa LinearAlgebra.BlasReal` is a malformed check and will always be false. We could change it to `eltype(samples) <: LinearAlgebra.BlasReal`, but the code within it calls a `dot` method that doesn't exist. Besides, for `StridedArray` inputs with `BlasReal` eltype, the second call to `dot` hits https://github.com/JuliaLang/julia/blob/v1.6.1/stdlib/LinearAlgebra/src/matmul.jl#L9, which forwards to `BLAS.dot` already, so this check is not necessary.

And it turns out removing the check gives us an almost 2x speed-up and cuts allocations substantially:

```julia
julia> using InferenceDiagnostics

julia> x = randn(1000, 10, 4);

julia> @btime ess_rhat($x);
```
current:
```julia
  222.542 μs (2759 allocations: 119.52 KiB)
```

this pr:
```julia
  140.919 μs (7 allocations: 31.95 KiB)
```

### `BDAESSMethod`:

Replacing a `sum` over a difference with 3 equivalent calls to `dot` enables us to get almost a 2x speed-up. The speed-up is probably due to hitting `BLAS.dot`.

```julia
julia> @btime ess_rhat($x; method=BDAESSMethod())
```

current:
```julia
  429.368 μs (56 allocations: 32.73 KiB)
```

this pr:
```julia
  220.347 μs (8 allocations: 31.98 KiB)
```


src/ess.jl
```diff
@@ -187,10 +178,11 @@ function mean_autocov(k::Int, cache::BDAESSCache)
     # compute mean autocovariance
     n = niter - k
     idxs = 1:n
+    idxs2 = k .+ idxs
```
devmotion, line 181: No need for broadcasting here:
```suggestion
    idxs2 = (k + 1):niter
```

src/ess.jl
```diff
@@ -187,10 +178,11 @@ function mean_autocov(k::Int, cache::BDAESSCache)
...
     s = Statistics.mean(1:nchains) do j
-        return sum(idxs) do i
-            abs2(samples[i, j] - samples[k + i, j])
-        end
+        s1 = view(samples, idxs, j)
+        s2 = view(samples, idxs2, j)
+        return LinearAlgebra.dot(s1, s1) + LinearAlgebra.dot(s2, s2) - 2 * LinearAlgebra.dot(s1, s2)
```
devmotion, line 185: I assume that in contrast to the previous implementation this can return negative values (similar to what I experienced with Distances: https://github.com/JuliaStats/Distances.jl/pull/161). So we should change this to
```suggestion
        r = LinearAlgebra.dot(s1, s1) + LinearAlgebra.dot(s2, s2) - 2 * LinearAlgebra.dot(s1, s2)
        return max(0, r)
```

In general, I am surprised that this is actually faster than the current implementation. In particular the allocations seem weird, there should be no difference here (or at least the current implementation should not be worse). I think this needs some more benchmarks and more thorough investigation (probably it would also be good to only benchmark this function and not `ess_rhat`).

src/ess.jl
```diff
@@ -144,16 +144,7 @@ function mean_autocov(k::Int, cache::ESSCache)
     firstrange = 1:(niter - k)
     lastrange = (k + 1):niter
     s = Statistics.mean(1:nchains) do i
-        if eltype(samples) isa LinearAlgebra.BlasReal
```
devmotion, line 147: Yes, this is incorrect and should be fixed. I think it would be good to benchmark the corrected version, in particular also with older Julia versions in which AFAIK `view`s were more expensive. The special `dot` call without views is e.g. also used in StatsBase: https://github.com/JuliaStats/StatsBase.jl/blob/2faa6e80b7966b915086d8cd5a4a4d89a2126db5/src/signalcorr.jl#L46 (where it was copied from I assume).

devmotion (changes requested):

src/ess.jl
sethaxen: > The special `dot` call without views is e.g. also used in StatsBase: https://github.com/JuliaStats/StatsBase.jl/blob/2faa6e80b7966b915086d8cd5a4a4d89a2126db5/src/signalcorr.jl#L46 (where it was copied from I assume).

It's not quite the same. In StatsBase, an `AbstractVector` is passed to `dot` This versions passes an `AbstractMatrix`, but it errors because there is no corresponding method. I don't think this branch ever worked, and I'm not certain how best to correct it for benchmarking.

sethaxen:
## Updated benchmarks

#### Setup

```julia
using InferenceDiagnostics, BenchmarkTools, Random
rng = MersenneTwister(42)
samples = randn(rng, 1000, 8)
chain_var = rand(rng, 8)
esscache = InferenceDiagnostics.ESSCache(samples, chain_var)
bdaesscache = InferenceDiagnostics.BDAESSCache(samples, chain_var, 0.0)
```

### `ESSMethod`

```julia
julia> @btime $(InferenceDiagnostics.mean_autocov)(1, $esscache);
```

#### Julia 1.0

current:
```julia
  2.529 μs (42 allocations: 1.52 KiB)
```

this pr:
```julia
  1.686 μs (17 allocations: 1.05 KiB)
```

#### Julia 1.6

current:
```julia
  3.256 μs (53 allocations: 1.78 KiB)
```

this pr:
```julia
  1.765 μs (0 allocations: 0 bytes)
```

### `ESSMethod`

```julia
julia> @btime $(InferenceDiagnostics.mean_autocov)(1, $bdaesscache);
```

#### Julia 1.6

current:
```julia
  12.282 μs (1 allocation: 16 bytes)
```

this pr:
```julia
  1.992 μs (1 allocation: 16 bytes)
```

src/ess.jl
sethaxen: I don't know where the allocations come from. But I realized just marking the inner loop of `sum` with `@inbounds` actually gave us more of a speed-up than using `dot`.

src/ess.jl
devmotion: Ah I missed that it also incorrectly used `samples` instead of `view(samples, :, i)`. But in this case one already creates a view 🤷🏻‍♂️ Of course, one could linearize the matrix with `vec` and update the ranges in every step but it seems unnecessarily complicated. Let's just drop the branch, since it hasn't been used there can't be any regressions anyway 👍

devmotion (approved):

sethaxen:
Hmm, isn't the format workflow supposed to review with suggested changes instead of just failing? I'll format locally. Btw, @devmotion would you mind adding me as a collaborator on this repo?

devmotion:
It opens suggestions only if the PR is not from a fork (due to Github's security restrictions for actions), and prints the suggestions in the logs of the action otherwise.

devmotion:
It seems recent changes in MLJ (refactoring of (MLJ)ScientificTypes) broke the MLJ integration.

Issue #2, opened 2021-06-13
Package name 


yebai:
`InferenceDiagnostics` sounds rather generic IMO. Does it make sense to narrow down a bit, e.g. `MCMCDiagnostics` or `MCMCTools`?

yebai:
Just found the following package with the name `MCMCDiagnostics`. Maybe we can join forces, merge the code and move the new library under the `ArviZ` umbrella? 

https://github.com/tpapp/MCMCDiagnostics.jl

devmotion:
The name was suggested and motivated  in https://github.com/TuringLang/MCMCChains.jl/issues/266 (the discussion also mentioned MCMCDiagnostics).

devmotion:
BTW this package (ie the code in MCMCChains) contains strictly more general implementations of the statistics in MCMCDiagnostics (eg three different algorithms for estimating the ESS).

yebai:
I'd prefer merging all the code (i.e. diagnostics code in MCMCChains, and MCMCDiagnostics)  into `MCMCDiagnostics`, then co-maintain the library under `ArviZ`. Such a plan would benefit all `DynamicHMC`, `Turing` and `ArviZ` users.

devmotion:
The intention of this package is to contain all the diagnostic code of MCMCChains. I don't think there's anything in MCMCDiagnostics that is missing and could be merged.

yebai:
I see - let's still make an effort at merging these packages to reduce redundancy in the Julia MCMC ecosystem. If there is no code to merge, we can simply ask the current `MCMCDiagnostics` author whether it's ok to transfer the package to `ArviZ` for (hopefully better) development.

sethaxen:
The motivation for `InferenceDiagnostics` is that we may want to include more general diagnostics than those for MCMC, e.g. for VI. One could create a `VIDiagnostics`, but IIUC there's enough overlapping code that it would make sense to have a single library. That being said, we could always start with `MCMCDiagnostics` and rename later if we do expand. Currently all diagnostics in the ArviZ package are MCMC-specific.

RE merging with `MCMCDiagnostics`, I agree that the diagnostics in this package are a superset of those in `MCMCDiagnostics` and that we should at least discuss a merge to the author.

devmotion:
How should we proceed? It would be nice to be able to finish https://github.com/TuringLang/MCMCChains.jl/pull/310 and to clean MCMCChains, so it would be good to start the registration process.

[...]

PR #16, opened 2021-07-04
Add information on methods for mcse

rikhuijzer:


src/mcse.jl
```diff
@@ -1,9 +1,13 @@
-#################### Monte Carlo Standard Errors ####################
-
 """
-    mcse(x::AbstractVector{<:Real}; method::Symbol=:imse, kwargs...)
+    mcse(x::AbstractVector{<:Real}, method::Symbol=:imse; kwargs...)
+
+Return Monte Carlo Standard Errors of samples `x`.
```
devmotion, line 4: Maybe keep
```suggestion
Compute the Monte Carlo standard error (MCSE) of samples `x`.
```
since it was supposed to improve the existing docstring in MCMCChains.

src/mcse.jl
```diff
@@ -1,9 +1,13 @@
...
+Here, `method` describes how to estimate the errors; possible options are:
+
```
devmotion, line 6:
```suggestion

The optional argument `method` describes how the errors are estimated. Possible options are: 
```

src/mcse.jl
```diff
@@ -1,9 +1,13 @@
...
+- `:bm` for batch means
+- `:imse` for the integrated mean-squared (prediction) error
+- `:ipse` for the initial positive sequence estimator
```
devmotion, line 9: At some point we should explain it more clearly but this is definitely an improvement :slightly_smiling_face:

src/mcse.jl
```diff
@@ -1,9 +1,13 @@
...
```
devmotion, line 10: It seems this line can be removed:
```suggestion
```

rikhuijzer:
Thanks for the feedback, David. Let me know if I can also update the MCMCChains docs. I'm not sure what you want with that exactly. For me, closing that PR is also fine if you plan to merge the migration to InferenceDiagnostics.jl soon(ish).

rikhuijzer:
According to Benjamin Deonovic on Slack. The methods are based on other calculations. Don't merge yet, I'll add refs.

rikhuijzer:
Okay. Updated it. Are the refs okay like this? I see that you also use 

```
**references**

[...]
```
but that isn't so useful here. Alternatively, we can move the docs in the different methods and add a bunch of `[mcse_bm](@ref)` etc

devmotion:
I think it's fine as it is. IIRC I did not use cross references since I was not satisfied with the Documenter output (e.g., there are issues with references that appear in different docstrings (https://github.com/JuliaDocs/Documenter.jl/issues/566) but maybe this is not an issue anymore since I removed the references from `FFTESSMethod`).

devmotion (approved):

devmotion:
> Thanks for the feedback, David. Let me know if I can also update the MCMCChains docs. I'm not sure what you want with that exactly. For me, closing that PR is also fine if you plan to merge the migration to InferenceDiagnostics.jl soon(ish).

I plan to register the package within the next days (https://github.com/devmotion/InferenceDiagnostics.jl/issues/2#issuecomment-873638740) and there is already a PR for MCMCChains, so I assume the migration can be completed until the end of next week. So maybe we can just wait until end of next week and merge the PR to MCMCChains only if it is not fixed until then?

rikhuijzer:
> > Thanks for the feedback, David. Let me know if I can also update the MCMCChains docs. I'm not sure what you want with that exactly. For me, closing that PR is also fine if you plan to merge the migration to InferenceDiagnostics.jl soon(ish).
> 
> I plan to register the package within the next days ([#2 (comment)](https://github.com/devmotion/InferenceDiagnostics.jl/issues/2#issuecomment-873638740)) and there is already a PR for MCMCChains, so I assume the migration can be completed until the end of next week. So maybe we can just wait until end of next week and merge the PR to MCMCChains only if it is not fixed until then?

Completely fine by me :) I'll just close the other PR for now.

devmotion:
Thanks for the PR!

PR #31, opened 2021-12-25
Update to MLJBase 0.19

devmotion:
Fixes #29. Fixes #30.


src/rstar.jl
```diff
@@ -112,7 +112,7 @@ function rstar(
 end
 
 # R⋆ for deterministic predictions (algorithm 1)
-function _rstar(predictions::AbstractVector, ytest::AbstractVector)
+function _rstar(predictions::AbstractVector{T}, ytest::AbstractVector{T}) where {T}
```
devmotion, line 115: The `_rstar` function signatures (see also below) have to be changed since MLJBase 0.19 uses probabilistic predictions that are not subtypes of `Distributions.Distribution`. MCMCChains downstream tests (performed with MLJBase 0.18) ensure that this change is non-breaking.

PR #33, opened 2022-04-01
Make Julia 1.6 the minimum supported version

sethaxen:
Since the `gammaccdf` function we use (via Distributions) from StatsFuns does not work on Julia v1.0 (https://github.com/JuliaStats/StatsFuns.jl/issues/136), our CI on 1.0 will error from now on. This PR drops support for older Julia versions, so that our minimum supported version is the latest LTS.


rikhuijzer (approved):
Great stuff. Let's see what David thinks.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "MCMCDiagnosticTools"
 uuid = "be115224-59cd-429b-ad48-344e309966f0"
 authors = ["David Widmann"]
-version = "0.1.3"
+version = "0.1.4"
```
rikhuijzer, line 4: Not allowed to bump Julia version in patch releases.

devmotion:
> Since the gammaccdf function we use (via Distributions) from StatsFuns does not work on Julia v1.0 (https://github.com/JuliaStats/StatsFuns.jl/issues/136), our CI on 1.0 will error from now on.

IMO this should be fixed upstream - if possible. `gammaccdf` is only problematic in the recent Rmath-reduced release of StatsFuns, versions < 0.9.16, (should) work just fine since they use Rmath instead of SpecialFunctions for these functions.

Project.toml
sethaxen: Is this documented somewhere? I didn't see this in https://pkgdocs.julialang.org/v1/compatibility/. ColPrac makes a good argument though for incrementing the minor version, but it's not required: https://github.com/SciML/ColPrac#dropping-support-for-earlier-versions-of-julia

It doesn't seem to be enforced, since I recently released a patch that did this: https://github.com/JuliaRegistries/General/pull/56946

devmotion:
Generally, I'm not happy about all the unrelated changes here. There's no need to change keyword argument syntax when updating the Julia compat to 1.6.

sethaxen:
> > Since the gammaccdf function we use (via Distributions) from StatsFuns does not work on Julia v1.0 ([JuliaStats/StatsFuns.jl#136](https://github.com/JuliaStats/StatsFuns.jl/issues/136)), our CI on 1.0 will error from now on.
> 
> IMO this should be fixed upstream - if possible. `gammaccdf` is only problematic in the recent Rmath-reduced release of StatsFuns, versions < 0.9.16, (should) work just fine since they use Rmath instead of SpecialFunctions for these functions.

How would you suggest fixing this upstream? Besides the approach suggested in https://github.com/JuliaStats/StatsFuns.jl/issues/136#issue-1189340686.

Project.toml
devmotion: It's documented here: https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#New-versions-of-existing-packages

devmotion:
> > > Since the gammaccdf function we use (via Distributions) from StatsFuns does not work on Julia v1.0 ([JuliaStats/StatsFuns.jl#136](https://github.com/JuliaStats/StatsFuns.jl/issues/136)), our CI on 1.0 will error from now on.
> > 
> > 
> > IMO this should be fixed upstream - if possible. `gammaccdf` is only problematic in the recent Rmath-reduced release of StatsFuns, versions < 0.9.16, (should) work just fine since they use Rmath instead of SpecialFunctions for these functions.
> 
> How would you suggest fixing this upstream? Besides the approach suggested in [JuliaStats/StatsFuns.jl#136 (comment)](https://github.com/JuliaStats/StatsFuns.jl/issues/136#issue-1189340686).

That doesn't fix the problem. These old Julia versions should just not use the non-Rmath versions because we can't fix the bugs for them. I.e., it should be fixed in the registry by changing Julia compat for the latest release.

sethaxen:
In that case I will close this PR and leave the fixing of this issue to someone else who knows which releases of StatsFuns should not have had Julia 1.0 compatibility.

Issue #10, opened 2021-06-16
Ensure all ArviZ diagnostics included here


sethaxen:
Companion to #4, we also should ensure all [diagnostic variations](https://arviz-devs.github.io/arviz/api/diagnostics.html) included in ArviZ are represented here. This issue will track that progress:

- [x] `bfmi` (#42)
- `ess`:
  - [x] `bulk`
  - [x] `tail`
  - [x] `quantile`
  - [x] `mean`
  - [x] `std`
  - [x] `median`
  - [x] `mad`
  - [ ] ~`z_scale`~
  - [ ] ~`folded`~
  - [ ] ~`identity`~
  - [ ] ~`local`~
- `mcse`:
  - [x] `mean`
  - [x] `std`
  - [x] `median`
  - [x] `quantile`
- `rhat`:
  - ~[ ] `rank`~
  - [x] `split`
  - [x] `bulk`
  - [x] `tail`
  - [ ] ~`folded`~
  - [ ] ~`z_scale`~
  - [ ] ~`identity`~

sethaxen:
This is now completed.

PR #42, opened 2022-08-09
Add bayesian fraction of missing information

sethaxen:
This PR adds a function `bfmi` for computing the estimated Bayesian fraction of missing information, which is a useful diagnostic for variants of HMC. AdvancedHMC and DynamicHMC both implement this diagnostic (as `EBFMI`). Even though it is specific to HMC, I include it here because:
1. transferring it here is part of #10
2. when called from a PPL, neither of the two HMC packages executes the diagnostic, and the diagnostic functions cannot in general be called on the outputs of PPLs. Generally a user will need to call the diagnostic themselves, and defining it here supports that.

The documentation of the function is adapted from ArviZ's. I've opted for `bfmi` instead of `EBFMI` as the name because
1. the fact this is an estimate is implied
2. lower case function names are more consistent with Julia conventions
3. this is the name ArviZ uses

While we could implement the function with fewer allocations, it's quite cheap already, and doing it this way allows for the possibility of preserving dimensions in the outputs if dimensional quantities are provided. Here's an example using ArviZ, which internally uses DimensionalData:

````julia
julia> using ArviZ

julia> idata = load_arviz_data("centered_eight")
InferenceData with groups:
  > posterior
  > posterior_predictive
  > sample_stats
  > prior
  > observed_data

julia> energy = idata.sample_stats.energy;

julia> bfmi(energy; dims=2)
4-element DimArray{Float64,1} with dimensions: 
  Dim{:chain} Sampled 0:3 ForwardOrdered Regular Points
and reference dimensions: 
  Dim{:draw} Sampled 249.0:499.0:249.0 ForwardOrdered Regular Points
 0.297861
 0.437247
 0.400779
 0.234243
````

MCMCChains uses AxisArrays internally, which often loses dimensions due to missing overloads. In this case, we lose dimensions due to no overload for `diff`:

````julia
julia> using Turing

julia> @model function model()
           x ~ filldist(Normal(), 10)
       end;

julia> chns = sample(model(), NUTS(), MCMCThreads(), 1_000, 4);

julia> energy = get(chns, :hamiltonian_energy).hamiltonian_energy;

julia> bfmi(energy)
4-element Vector{Float64}:
 1.152615860347193
 1.1015170772812852
 1.1757399849528423
 1.1924088595379851
````

### CODE REVIEW

File: src/bfmi.jl [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+    A Conceptual Introduction to Hamiltonian Monte Carlo.
+    [arXiv:1701.02434v2](https://arxiv.org/pdf/1701.02434v2.pdf) [stat.ME]
+[^Betancourt2016]: Betancourt M. (2016).
+    Diagnosing Suboptimal Cotangent Disintegrations in Hamiltonian Monte Carlo.
+    [arXiv:1604.00695v1](https://arxiv.org/pdf/1604.00695v1.pdf) [stat.ME]
+"""
+function bfmi(energy; dims=1)
+    energy_diff = diff(energy; dims)
+    energy_var = Statistics.var(energy; dims)
+    result = dropdims(Statistics.mean(abs2, energy_diff; dims) ./ energy_var; dims)
```
devmotion, line 39:
```suggestion
    energy_diff = diff(energy; dims=dims)
    energy_var = Statistics.var(energy; dims=dims)
    result = dropdims(Statistics.mean(abs2, energy_diff; dims=dims) ./ energy_var; dims=dims)
```

src/bfmi.jl
```diff
@@ -0,0 +1,41 @@
...
+# References
+
+[^Betancourt2018]: Betancourt M. (2018).
+    A Conceptual Introduction to Hamiltonian Monte Carlo.
+    [arXiv:1701.02434v2](https://arxiv.org/pdf/1701.02434v2.pdf) [stat.ME]
+[^Betancourt2016]: Betancourt M. (2016).
+    Diagnosing Suboptimal Cotangent Disintegrations in Hamiltonian Monte Carlo.
+    [arXiv:1604.00695v1](https://arxiv.org/pdf/1604.00695v1.pdf) [stat.ME]
+"""
+function bfmi(energy; dims=1)
```
devmotion, line 36: Should this be restricted to something like
```suggestion
function bfmi(energy::AbstractArray{<:Real}; dims=1)
```
as other methods?

src/bfmi.jl
```diff
@@ -0,0 +1,41 @@
...
+    energy_diff = diff(energy; dims)
+    energy_var = Statistics.var(energy; dims)
+    result = dropdims(Statistics.mean(abs2, energy_diff; dims) ./ energy_var; dims)
+    return iszero(ndims(result)) ? result[] : result
```
devmotion, line 40: Doesn't this break type inference (in some cases at least, e.g. older Julia versions)? Would it be safer to check e.g.
```suggestion
    return result isa AbstractArray{<:Any,0} ? result[] : result
```
?

src/bfmi.jl
devmotion: And do we want to support tuples, arrays, etc. for `dims` or rather restrict it to `dims::Int`?

test/bfmi.jl
```diff
@@ -0,0 +1,41 @@
+@testset "bfmi.jl" begin
+    energy = [1, 2, 3, 4]
+    @test @inferred(bfmi(energy)) isa Real
+    @test bfmi(energy) ≈ 0.6
```
devmotion, line 4: How are these reference values computed? Maybe add a comment?

src/bfmi.jl
sethaxen: > Should this be restricted to something like
> 
> as other methods?

For consistency, probably.



> And do we want to support tuples, arrays, etc. for `dims` or rather restrict it to `dims::Int`?

I don't see a use case for tuple or array dims (inputs must either be a vector of draws or a matrix of draws for multiple chains). However, I don't think we should restrict to `Int` dims. e.g. for dimensional arrays, we can pass a specialized dimension object to select a draw dimension by name. e.g. from the ArviZ example:

```julia
julia> bfmi(energy; dims=:draw)
4-element DimArray{Float64,1} with dimensions: 
  Dim{:chain} Sampled 0:3 ForwardOrdered Regular Points
and reference dimensions: 
  Dim{:draw} Sampled 249.5:500.0:249.5 ForwardOrdered Regular Points
 0.289451
 0.409332
 0.4406
 0.235864

julia> bfmi(permutedims(energy); dims=:draw)
4-element DimArray{Float64,1} with dimensions: 
  Dim{:chain} Sampled 0:3 ForwardOrdered Regular Points
and reference dimensions: 
  Dim{:draw} Sampled 249.5:500.0:249.5 ForwardOrdered Regular Points
 0.289451
 0.409332
 0.4406
 0.235864
```

src/bfmi.jl
sethaxen: Not certain, but I just ran on Julia 1.0, and `dims` is not supported as a kwarg if `energy` is a vector, so I'll split them into 2 methods.

test/bfmi.jl
sethaxen: Sure, they're computed by comparison with Python ArviZ. Will add a note.

src/bfmi.jl
devmotion: > However, I don't think we should restrict to `Int` dims. e.g. for dimensional arrays, we can pass a specialized dimension object to select a draw dimension by name. e.g. from the ArviZ example:

Wouldn't we always want to use `:draw` in that example? I.e., wouldn't it be better to add a dispatch for `DimArray` that always picks `:draw` (at least if the dimensions contain `Dim{:draw}` which - I assume, not familiar with that package - could be determined at compile time) instead of supporting arbitrary `dims`?

sethaxen: If `DimArray` was owned by ArviZ, yes, but since it isn't, that would be type piracy. e.g. someone else could use DimensionalData to store MCMC draws independently of ArviZ and name the draw dimension `:samples`. In practice, we would define a method like:

```julia
function MCMCDiagnosticTools.bfmi(idata::InferenceData)
    energy = idata.sample_stats.energy
    dims = Dimensions.dims(energy, :draw)
    MCMCDiagnosticTools.bfmi(energy; dims)
end
```
which we can do in ArviZ because `:draw` is a reserved name for MCMC draw dimensions. Another perfectly valid option uses `Dimensions.dimnum(energy, :draw)` to get the integer dimension. Both resolve at compile time (I assume due to inlining).

What problems do you see with supporting arbitrary `dims`? We're just forwarding them to other methods, so if other methods support the provided dims, then so would we. This seems to me to be preferable. By not setting type constraints, we allow for more general and composable code. e.g. `LogExpFunctions.logsumexp` just works for DimensionalData since it puts no constraints on `dims` and instead just forwards them to `reduce`.

devmotion: My main concern here is that in my experience unrestricted `dims` keyword arguments usually/often also support tuples, arrays, etc. of dimensions but we don't want that here (in contrast to e.g. `logsumexp`, and therefore I think the comparison is a bit flawed). The simplest way to avoid these issues would be to just restrict `dims` to `dims::Int` for now. That seems to work for the usual inputs in MCMCDiagnosticTools (regular arrays) and would also work for structures such as `InferenceData` or `MCMCChains.Chain` by forwarding calls to the underlying wrapped array (for which `dims::Int` would be sufficient)?

sethaxen: While I disagree that this is a problem here (we don't use `dims`, we just forward it and let other methods error if what is requested is unreasonable), it's true that in ArviZ and MCMCChains (I assume) we can make do with an integer constraint, and I'm okay with this compromise.

devmotion: Somehow I tend towards the more conservative approach here and think `dims::Int` would be a good start. We can easily remove the type annotation if it turns out that it is too restrictive.

sethaxen: Done!

devmotion (approved):

Issue #49, opened 2022-11-18
Changes to dimension ordering


sethaxen:
Following the discussion in #5, I propose the following dimension interpretations be used uniformly across the package.
- `AbstractVector`: `(draws,)`, vector of MC draws for a single parameter
- `AbstractArray{<:Any,3}`: `(params, draws, chains)`, array of MCMC draws for multiple parameters

The first is obvious and consistent with the current interpretation. The second is consistent with Julia's default column ordering, as I've explained in https://github.com/arviz-devs/InferenceObjects.jl/issues/8.

I also propose that all of our common diagnostics ultimately implement methods for the `AbstractArray{<:Any,3}` signature. This allows the user to try multiple diagnostics on a single array format without needing to do different reshapes and slices for each diagnostic, but it doesn't require users use `AbstractArray{<:Any,3}` if their draws are not in that format.

When outputs contain a subset of these dimensions, they should preserve the order of the dimensions. I don't think this requires any changes right now.

## Concrete changes

Breaking:
- `discretediag`, `ess`, `gelmandiag`: `(draws, params, chains)` -> `(params, draws, chains)`

New methods:
- `rstar`
  - change sample matrix input to have shape `(params, draws)` instead of `(draws, params)`
  - add method with `(params, draws, chains)` input, which is then forwarded to the current method after a `reshape` and `repeat` (to generate chain indices)
- `mcse`: add an `AbstractArray{<:Any,3}` method.

Things not changed:
- `bfmi`: it makes no sense for `energy` to have a `params` dimension, so `bfmi` does not need a 3d array method
- `gewekediag`, `heideldiag`, `rafterydiag`: I suspect these are rarely used, so adding `AbstractArray{<:Any,3}` methods would be low priority

cc @devmotion @ParadaCarleton

devmotion:
I guess the main drawback is that we would have to use `permutedims` in MCMCChains which currently uses the layout `(draws, params, chains)`. It would be good to check if/how that impacts performance of the analysis there. I assume if we manage to reduce the amount of `permutedims` by not re-doing it for every analysis when evaluating multiple methods then it should be fine.

sethaxen:
[...]

Issue #39, opened 2022-05-19
Improved MCSE


sethaxen:
See corresponding issues in posterior (https://github.com/stan-dev/posterior/issues/232) and ArviZ (https://github.com/arviz-devs/arviz/issues/2005)

sethaxen:
More broadly, we should update our implementations of `mcse` in several ways:
- accept 3D arrays as inputs, where the estimator takes into account between-chain convergence. (e.g. the `imse` method for a single chain should yield a similar estimate to using `ess_rhat` for a single chain without splitting. But a better estimate with poor mixing comes from using the ESS estimate)
- accept an optional `estimator` argument, like `ess_rhat` in #53. Estimators have been worked out for several functions (`mean`, `std`, `quantile`, see [e.g. posterior: https://github.com/stan-dev/posterior/blob/879bb0f72f67dfb67637bac2d05abcc1fa173a3b/R/convergence.R](https://arxiv.org/abs/1903.08008))
- update batch-means approach to work for any estimator
- add subsampling bootstrap method (see https://github.com/arviz-devs/arviz/pull/1974)

I'm not 100% certain what the API should look like. My inclination is that we should ultimately move away from our existing implementations, since they ignore between-chain convergence and so likely underestimate the MCSE. The ESS approach should be used by default for the estimators that use it (accepting a method keyword forwarded to `ess_rhat`), and this replaces `imse`. I'm not certain how useful `ipse` is. For the fallback for generic estimators, we should benchmark bootstrapping vs batch-means.

Issue #22, opened 2021-09-26
Add rank-normalized ESS and other variants


sethaxen:
ESS is defined in the context of a specific estimate. For example, our current ESS implementations are all in the context of estimating the mean, so they will have problems whenever the mean/variance is not finite, e.g. for the Cauchy distribution. [Vehtari et al](https://arxiv.org/pdf/1903.08008.pdf) proposed several variants to the ESS/R-hat for diagnosing various problems that can manifest in posterior draws.

The different variants covered in the paper fall in the following categories

Splitting:
- split
- non-split

Pre-processing:
- rank-normalization to satisfy assumption of normality
- folding to avoid R-hat being fooled when chains have same location but different scales

Quantity being estimated:
- mean
- std
- quantile
- quantile interval
- mad

Some notes:
- splitting seems universally better than non-splitting, so I don't think we should support non-splitting.
- rank-normalization seems in general to be more robust than non-normalization, except for when one computes an MCSE; then one needs the un-rank-normalized version. So to me it makes sense for bulk-ESS (rank-normalized, unfolded ESS for estimate of mean) to be the default ESS method, which would also be consistent with ArviZ and Stan.
- Estimating ESSs for specific estimates generally requires pre- and post-processing of ESS estimates. Some examples given below (ignore the methods, this is not an API proposal):

```julia
ess_rhat(x, ::Val{:bulk}; kwargs...) = ess_rhat(rank_normalize(x); kwargs...)
function ess_rhat(x, ::Val{:tail}, p=0.90; kwargs...)
    return min(ess_rhat(x, Val(:quantile), (1-p)/2; kwargs...), ess_rhat(x, Val(:quantile), (1+p)/2; kwargs...))
end
ess_rhat(x, ::Val{:fold}; kwargs...) = ess_rhat(rank_normalize(fold(x)))
ess_rhat(x, ::Val{:quantile}, p; kwargs...) = ess_rhat(rank_normalize(qindicator(x, p)); kwargs...)
ess_rhat(x, ::Val{:median}; kwargs...) = ess_rhat(x, Val(:quantile), 0.5; kwargs...)
ess_rhat(x, ::Val{:mad}; kwargs...) = ess_rhat(rank_normalize(fold(x)), Val(:median); kwargs...)
function ess_rhat(x, ::Val{:qinterval}, pl, pu; kwargs...)
    return ess_rhat(rank_normalize(qindicatorrange(x, pl, pu)); kwargs...)
[...]

PR #50, opened 2022-11-19
Change dimension ordering

sethaxen:
Changes dimension ordering for 3d arrays to be ~`(nparams, ndraws, nchains)`~ `(ndraws, nchains, nparams)` and ~adds method with 3d array inputs for `mcse`~ as proposed in #49.


src/discretediag.jl
```diff
@@ -441,9 +443,9 @@ function discretediag(
     )
     0 < frac < 1 || throw(ArgumentError("`frac` must be in (0,1)"))
 
-    num_iters = size(chains, 1)
+    num_iters = size(chains, 2)
     between_chain_vals, within_chain_vals, _, _ = discretediag_sub(
-        chains, frac, method, nsim, num_iters, num_iters
+        permutedims(chains, (2, 1, 3)), frac, method, nsim, num_iters, num_iters
     )
```
devmotion, line 449: Is it possible to change `discretediag_sub` such that we can avoid the `permutedims` call here?

src/ess.jl
```diff
@@ -238,7 +238,7 @@ function ess_rhat(
     rhat = Vector{T}(undef, nparams)
 
     # for each parameter
-    for (i, chains_slice) in enumerate((view(chains, :, i, :) for i in axes(chains, 2)))
+    for (i, chains_slice) in enumerate((selectdim(chains, 1, i) for i in axes(chains, 1)))
```
devmotion, line 241: Isn't that just
```suggestion
    for (i, chains_slice) in enumerate(eachslice(chains; dims=1))
```
?

src/gelmandiag.jl
```diff
@@ -1,14 +1,14 @@
 function _gelmandiag(psi::AbstractArray{<:Real,3}; alpha::Real=0.05)
-    niters, nparams, nchains = size(psi)
+    nparams, niters, nchains = size(psi)
     nchains > 1 || error("Gelman diagnostic requires at least 2 chains")
 
     rfixed = (niters - 1) / niters
     rrandomscale = (nchains + 1) / (nchains * niters)
 
-    S2 = map(Statistics.cov, (view(psi, :, :, i) for i in axes(psi, 3)))
+    S2 = map(x -> Statistics.cov(x; dims=2), (view(psi, :, :, i) for i in axes(psi, 3)))
```
devmotion, line 8:
```suggestion
    S2 = map(x -> Statistics.cov(x; dims=2), eachslice(psi; dims=3))
```
?

src/mcse.jl
```diff
@@ -23,6 +25,9 @@ function mcse(x::AbstractVector{<:Real}; method::Symbol=:imse, kwargs...)
         throw(ArgumentError("unsupported MCSE method $method"))
     end
 end
+function mcse(x::AbstractArray{<:Real,3}; kwargs...)
+    return dropdims(mapslices(xi -> mcse(vec(xi); kwargs...), x; dims=(2, 3)); dims=(2, 3))
```
devmotion, line 29: IIRC `mapslices` is quite inefficient and usually not very performant. I wonder if there's a performance problem here and, if yes, if it could be fixed?
Maybe also just use
```suggestion
    return vec(mapslices(xi -> mcse(vec(xi); kwargs...), x; dims=(2, 3)))
```
?

src/rstar.jl
```diff
@@ -2,14 +2,16 @@
     rstar(
         rng=Random.GLOBAL_RNG,
         classifier,
-        samples::AbstractMatrix,
-        chain_indices::AbstractVector{Int};
+        samples::AbstractArray,
+        [chain_indices::AbstractVector{Int}];
```
devmotion, line 6: This seems a bit confusing: According to the docstring we only support matrices and 3-dim arrays not generic `AbstractArray`, and it seems `chain_indices` is not supported if `samples` is a 3-d array. Maybe the only clear documentation are separate docstrings (where in the text one could refer to the other)?

src/rstar.jl
```diff
@@ -88,11 +88,11 @@ function rstar(
     # train classifier on training data
     ycategorical = MLJModelInterface.categorical(y)
     fitresult, _ = MLJModelInterface.fit(
-        classifier, verbosity, Tables.table(x[train_ids, :]), ycategorical[train_ids]
+        classifier, verbosity, Tables.table(x[:, train_ids]'), ycategorical[train_ids]
```
devmotion, line 91: A bit unfortunate that the memory layout does not match the MLJ conventions anymore in this case. But probably unavoidable? Or should the new convention only apply to 3d arrays?

src/ess.jl
sethaxen: `eachslice` isn't present in Julia 1.0. We could add Compat as a direct dependency (it's already an indirect one) to use it.

src/mcse.jl
sethaxen: Probably. `mapslices` is also not type-inferrable. We might be more efficient by using `eachslices` and constructing the output container ourselves.

I prefer `dropdims` here instead of `vec` because it's more friendly with named dimensions. e.g. this currently works:

```julia
julia> using MCMCDiagnosticTools, DimensionalData

julia> da = DimArray(randn(3, 10, 4), (Dim{:var}(1:3), Dim{:draw}(1:10), Dim{:chain}(1:4)));

julia> mcse(da)
3-element DimArray{Float64,1} with dimensions: 
  Dim{:var} Sampled{Int64} 1:3 ForwardOrdered Regular Points
and reference dimensions: 
  Dim{:draw} Sampled{Int64} 1:1 ForwardOrdered Regular Points,
  Dim{:chain} Sampled{Int64} 1:1 ForwardOrdered Regular Points
 1  0.187351
 2  0.146984
 3  0.159
```
In general our functions won't preserve named dimensions, but when they trivially can, it makes sense to do so.

src/ess.jl
devmotion: If it's an indirect one anyway, then maybe let's use `eachslice` + Compat? I think that's better in the long run since I assume we won't remember to update these lines when we drop support for older Julia versions.

src/mcse.jl
devmotion: I see, yes, that's a valid argument against `vec`.

src/rstar.jl
sethaxen: It might be confusing to the user if the order of two dimensions was swapped when comparing the 2 methods.

Also, even though MLJ assumes columns are features, not every classifier does, so sometimes internally the table is transposed. (see https://alan-turing-institute.github.io/MLJ.jl/dev/quick_start_guide_to_adding_models/#Regressor)

I wonder if it makes sense to relax the constraint that `X` be a matrix input to this function and instead require just that it be table whose columns are parameters. This only makes sense if we can use MLJBase API functions to do the subsetting within the method. Then the 3d-array method would just construct the necessary Table.

src/discretediag.jl
sethaxen: Yes! I plan to, it's just a little more complicated.

src/rstar.jl
sethaxen: e.g. it looks like `partition` would work: https://juliaai.github.io/MLJBase.jl/stable/datasets/#MLJBase.partition-Tuple{Any,%20Vararg{Real}}

devmotion: We don't use MLJBase but only MLJModelInterface since that would pull in additional dependencies. There are some design issues in MLJ though it seems since MLJModelInterface is not completely consistent (see, e.g., the comment in the file and the internal workaround I had to add recently).

devmotion: But in principle, I like the idea of supporting generic tables. In that case it might be natural to even just make the chain index another column?

src/gelmandiag.jl
sethaxen: For some reason this one causes problems with type inference.

src/mcse.jl
sethaxen: I replaced this with an explicit allocation of the output container using the same type heuristic as `ess_rhat`, followed by a use of `eachslice` to fill.

src/gelmandiag.jl
sethaxen: For now I've reverted.

src/rstar.jl
sethaxen: I was able to support generic tables, `AbstractMatrix`, and `AbstractVector` inputs using just MLJModelInterface utilities.

> In that case it might be natural to even just make the chain index another column?

Potentially, but there are 2 reasons not to do this. First, whether constructing the inputs from a vector of matrices of draws from different chains or constructing from a 3d array, it's easy to make a table of predictors, but it's another step for the user to append the column of chain IDs to this. Second, MLJ's interface separates the two, and since one of the arguments is constrained to be an MLJ type, it makes sense to be consistent with MLJ's interface.

sethaxen:
I opted to wait to allow `mcse` to take 3d inputs until tackling #39 and #22. @devmotion could you re-review this?

sethaxen:
Bump @devmotion do you think you will have time to review this soon?

src/rstar.jl
```diff
@@ -115,14 +138,24 @@ function _predict(model::MLJModelInterface.Model, fitresult, x)
...
+    x::AbstractArray{<:Any,3};
     kwargs...,
 )
+    samples = reshape(x, :, size(x, 3))
+    chain_inds = repeat(axes(x, 2); inner=size(x, 1))
+    return rstar(rng, classifier, samples, chain_inds; kwargs...)
+end
+
+function rstar(classif::MLJModelInterface.Supervised, x, y::AbstractVector{Int}; kwargs...)
     return rstar(Random.GLOBAL_RNG, classif, x, y; kwargs...)
```
devmotion, line 152:
```suggestion
    return rstar(Random.default_rng(), classif, x, y; kwargs...)
```

src/rstar.jl
```diff
@@ -115,14 +138,24 @@ function _predict(model::MLJModelInterface.Model, fitresult, x)
...
 end
 
+function rstar(classif::MLJModelInterface.Supervised, x::AbstractArray{<:Any,3}; kwargs...)
+    return rstar(Random.GLOBAL_RNG, classif, x; kwargs...)
```
devmotion, line 156:
```suggestion
    return rstar(Random.default_rng(), classif, x; kwargs...)
```

src/gelmandiag.jl
```diff
@@ -1,14 +1,14 @@
 function _gelmandiag(psi::AbstractArray{<:Real,3}; alpha::Real=0.05)
-    niters, nparams, nchains = size(psi)
+    niters, nchains, nparams = size(psi)
     nchains > 1 || error("Gelman diagnostic requires at least 2 chains")
 
     rfixed = (niters - 1) / niters
     rrandomscale = (nchains + 1) / (nchains * niters)
 
-    S2 = map(Statistics.cov, (view(psi, :, :, i) for i in axes(psi, 3)))
+    S2 = map(x -> Statistics.cov(x; dims=1), (view(psi, :, i, :) for i in axes(psi, 2)))
```
devmotion, line 8:
```suggestion
    # `eachslice(psi; dims=2)` breaks type inference 
    S2 = map(x -> Statistics.cov(x; dims=1), (view(psi, :, i, :) for i in axes(psi, 2)))
```

src/rstar.jl
```diff
@@ -102,6 +122,9 @@ function rstar(
     return result
 end
 
+_astable(x::AbstractVecOrMat) = Tables.table(x)
```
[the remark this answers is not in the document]
sethaxen: There's some, but it's unlikely to be a bottleneck.

```julia
julia> using MLJBase, Tables, BenchmarkTools

julia> foo(x) = MLJBase.matrix(Tables.table(x));

julia> x = randn(4000, 100);

julia> @btime foo($x);
  21.959 μs (508 allocations: 30.00 KiB)

julia> x = randn(4000);

julia> @btime foo($x);
  298.245 ns (12 allocations: 936 bytes)
```

MLJBase documentation recommends using a Table input for `x` though, and then they will convert to a `Matrix` if the particular model requires it.

src/rstar.jl
```diff
@@ -1,15 +1,15 @@
 """
     rstar(
-        rng=Random.GLOBAL_RNG,
-        classifier,
-        samples::AbstractMatrix,
-        chain_indices::AbstractVector{Int};
+        rng::Random.AbstractRNG=Random.GLOBAL_RNG,
```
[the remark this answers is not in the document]
sethaxen: > But maybe that's not supported by Julia 1.0? In any case, AFAIK Random.GLOBAL_RNG only exists for backwards compatibility.

Ah good to know! The docstrings all say the default is `Random.GLOBAL_RNG`, but indeed, since 1.3.0 the implementations have all used `default_rng()`. I'm not opposed to dropping support for <1.6, though this would only require us to drop <1.3.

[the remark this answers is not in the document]
devmotion: If Julia 1.3 still works, IMO there's no reason to drop support for it :slightly_smiling_face: I wonder, would that allow us to remove the Compat dependency again?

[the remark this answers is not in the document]
sethaxen: Yes, we only need Compat to support v1.0. Even v1.1 would let us drop it. Although maybe this isn't even an issue now, because doesn't our `rstar` implementation already require v1.3? With our current version bounds, we can't even install our package and MLJBase in the same environment on v1.2.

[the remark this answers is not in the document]
devmotion: Well, it really depends on what model you want to use in `rstar`. Installing MCMCDiagnosticTools and MLJBase in the same environment works fine with e.g. Julia 1.0. But I'm not sure if there are any compatible MLJ classifiers. In the tests we also use MLJXGBoostInterface and that does not work on Julia < 1.3 (and also not on 32bit). But possibly some other classifier would work.

In any case, I think it would be reasonable to drop support for Julia < 1.3. Even more so since it will be a breaking release.

Project.toml
```diff
@@ -1,10 +1,11 @@
 name = "MCMCDiagnosticTools"
 uuid = "be115224-59cd-429b-ad48-344e309966f0"
 authors = ["David Widmann"]
-version = "0.1.5"
+version = "0.2.0"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
devmotion, line 8:
```suggestion
```

devmotion (approved):
Looks good to me, thanks!

PR #52, opened 2022-12-13
Implement fixes for rstar

### LINKED ISSUE #51: Proposed changes to rstar defaults
The defaults of our `rstar` differ slightly from the paper:
- while the paper recommends a 70-30 split of train-test data, we default to 80-20
- the paper describes the split as happening within each chain, i.e. it is stratified by chains, while we split all draws
- while the paper recommends splitting each chain in two to diagnose poor within-chain convergence as well, we do not offer this as an option (it would be straightforward with #50)

I propose that with the breaking changes in #50 we update `rstar` to:
- use chain-stratified splitting with an ~80-20~ 70-30 split by default
- add a keyword `split_chains`, defaulting to `true`.

sethaxen:
Fixes #51: 
- uses a 70-30 split for training/test data
- adds a keyword argument `nsplit=2` (as suggested in https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/22#issuecomment-1338251834) to control how many chains each individual chain is split into (`nsplit=1` is the old behavior) to check for within-chain convergence.
- uses stratified sampling to split, which ensures that `frac` draws for each chain are in the training data 

I'm not thrilled with the name `nsplit`, as it's not terribly descriptive, but I haven't thought of a better one that wasn't quite verbose.

As suggested in https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/51#issuecomment-1347262676, we consider these changes non-breaking because they make the defaults consistent with the recommendations in the paper.


devmotion:
> I'm not thrilled with the name nsplit, as it's not terribly descriptive, but I haven't thought of a better one that wasn't quite verbose.

Maybe we could use even just `split`? Or `split_chains`?

sethaxen:
> Maybe we could use even just `split`? Or `split_chains`?

What I don't like about these is that if the user forgets `split` or `split_chains` is a number of splits, they might set `split_chains=true` thinking its boolean, and accidentally disable splitting. At least `nsplit` conveys that it's the number of splits. Something like `num_chain_splits` is a little ugly but maybe less prone to mistakes?

devmotion:
> What I don't like about these is that if the user forgets split or split_chains is a number of splits, they might set split_chains=true thinking its boolean, and accidentally disable splitting. At least nsplit conveys that it's the number of splits. Something like num_chain_splits is a little ugly but maybe less prone to mistakes?

If we enforce `split::Int` the function will error if someone sets `split=true`:
```julia
julia> f(; split::Int=2) = split
f (generic function with 1 method)

julia> f()
2

julia> f(; split=4)
4

julia> f(; split=true)
ERROR: TypeError: in keyword argument split, expected Int64, got a value of type Bool
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```

src/utils.jl
```diff
@@ -0,0 +1,82 @@
...
+corresponding indices in `x`.
+"""
+function indices_of_unique(x)
+    d = Dict{eltype(x),Vector{Int}}()
+    for (i, xi) in enumerate(x)
+        if haskey(d, xi)
+            push!(d[xi], i)
+        else
+            d[xi] = [i]
+        end
```
devmotion, line 14: This can be made more efficient by not looking up the key twice. One could e.g. use
```suggestion
        d_xi = get!(d, xi) do
            return Int[]
        end
        push!(d_xi, i)
```

Apart from that, it seems like a function that could exist e.g. in StatsBase (similar to `proportionmap` etc.). Did you check that?

src/utils.jl
```diff
@@ -0,0 +1,82 @@
+"""
+    indices_of_unique(x) -> Dict
+
+Return a `Dict` whose keys are the unique elements of `x` and whose values are the
+corresponding indices in `x`.
+"""
+function indices_of_unique(x)
+    d = Dict{eltype(x),Vector{Int}}()
+    for (i, xi) in enumerate(x)
```
devmotion, line 9: This will only give you the indices of `x` if `x` actually uses linear indices of type `Int`. But it seems it will fail for more general indices. Maybe one could use
```suggestion
    for i in eachindex(x)
        xi = x[i]
```
instead (and adapt the value type of `d` accordingly)?

src/utils.jl
```diff
@@ -0,0 +1,82 @@
...
+            push!(d[xi], i)
+        else
+            d[xi] = [i]
+        end
+    end
+    return d
+end
+
+"""
+    split_chain_indices(
```
devmotion, line 20: Isn't there some existing splitting functionality for `ess`? Is the plan to merge these eventually?

src/utils.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+function split_chain_indices(c::AbstractVector{<:Int}, nsplit::Int=2)
+    cnew = similar(c)
+    if nsplit == 1
+        copyto!(cnew, c)
+        return cnew
+    end
+    chain_indices = indices_of_unique(c)
+    chain_ind = 0
+    for chain in sort(collect(keys(chain_indices)))
```
devmotion, line 39: You could avoid sorting if you directly use an `OrderedDict` I assume.

src/utils.jl
```diff
@@ -0,0 +1,82 @@
...
+        inds = chain_indices[chain]
+        ndraws_per_split, rem = divrem(length(inds), nsplit)
+        ilast = 0
+        for j in 1:nsplit
+            chain_ind += 1
+            ndraws_this_split = ndraws_per_split + (j ≤ rem)
+            i = ilast + 1
+            ilast = i + ndraws_this_split - 1
+            @views cnew[inds[i:ilast]] .= chain_ind
+        end
```
devmotion, line 49: Isn't there some partitioning function in base or Base.Iterators that we could use here?

src/utils.jl
```diff
@@ -0,0 +1,82 @@
...
+
+This is used, for example, to split data into training and test data while preserving the
+class balances.
+"""
+function shuffle_split_stratified(
+    rng::Random.AbstractRNG, groups::AbstractVector, frac::Real
+)
+    inds1 = Int[]
+    inds2 = Int[]
+    group_indices = indices_of_unique(groups)
```
devmotion, line 72: It seems also this function relies on the fact that indices are of type `Int`.

src/utils.jl
sethaxen: Not quite merge, because there are two different types of splitting we can consider. This approach supports ragged chains and is as a result more complex and doesn't discard any draws (instead dividing the remainder across the earlier splits).

For `ess`/`rhat`, we don't support ragged chains so would discard draws if necessary to keep them the same length after splitting. This implementation is much simpler and can be done in a non-allocating way with just `reshape` and `view` on a 3d array. This will be part of https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/22.

sethaxen: The existing splitting functionality `copy_split!` will go away.

src/utils.jl
sethaxen: Ah, of course!

src/utils.jl
devmotion: Sounds good, I just want to make sure we use existing functionality. If it doesn't exist yet that's unfortunate but, of course, then we should use our own implementation.

devmotion: Actually, it seems `indicatormap` returns the information we are interested in: https://juliastats.org/StatsBase.jl/stable/misc/#StatsBase.indicatormat But maybe it's not the desired output format for our purposes.

sethaxen: It's similar, yes, but a little clunky. e.g. here's how we could get the vector of indices:
```julia
using SparseArrays
map(first ∘ findnz ∘ sparse, eachslice(indicatormat(x; sparse=true); dims=1))
```
But I still think it makes more sense to try to upstream the functionality we want, since often something like what we want will be more convenient for the user.

devmotion: Yes, I agree. That seems a bit inconvenient.

src/utils.jl
sethaxen: It turns out we can't use `Iterators.partition` because it only lets us specify the partition size, not the number of partitions. e.g. suppose we want to split `1:4` into 3 partitions, `Iterators.partition(1:4, 2)` gives us `[[1, 2], [3, 4]]`, and `Iterators.partition(1:4, 1)` gives us `[[1], [2], [3], [4]]`, but we want `[[1, 2], [3], [4]]`. I don't see another iterator in `Base.Iterators` that would work for us.

devmotion: Ah I see. Thanks for investigating it though.

src/utils.jl
```diff
@@ -0,0 +1,96 @@
...
+group are in `inds1` and the remainder are in `inds2`.
+
+This is used, for example, to split data into training and test data while preserving the
+class balances.
+"""
+function shuffle_split_stratified(
+    rng::Random.AbstractRNG, groups_ids::AbstractVector, frac::Real
+)
+    T = eltype(eachindex(groups_ids))
+    groups, indices = unique_indices(groups_ids)
```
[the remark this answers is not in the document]
sethaxen: Sure, but these produce different results. We want the eltypes of the elements of `indices`, not `indices` itself.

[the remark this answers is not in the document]
devmotion: Ah true, it should be
```suggestion
    groups, indices = unique_indices(groups_ids)
    T = eltype(eltype(indices))
```

src/utils.jl
```diff
@@ -0,0 +1,96 @@
...
+    rng::Random.AbstractRNG, groups_ids::AbstractVector, frac::Real
+)
+    groups, indices = unique_indices(groups_ids)
+    T = eltype(eltype(indices))
+    N1_tot = sum(x -> round(Int, length(x) * frac), indices)
+    N2_tot = length(groups_ids) - N1_tot
+    inds1 = Vector{T}(undef, N1_tot)
+    inds2 = Vector{T}(undef, N2_tot)
+    items_in_1 = items_in_2 = 0
+    for (group, inds) in zip(groups, indices)
```
devmotion, line 85: We never use `group` anywhere it seems?
```suggestion
    _, indices = unique_indices(groups_ids)
    T = eltype(eltype(indices))
    N1_tot = sum(x -> round(Int, length(x) * frac), indices)
    N2_tot = length(groups_ids) - N1_tot
    inds1 = Vector{T}(undef, N1_tot)
    inds2 = Vector{T}(undef, N2_tot)
    items_in_1 = items_in_2 = 0
    for inds in indices
```

sethaxen:
@devmotion I have implemented all of your suggestions

The failed integration test with MCMCChains is expected. That particular failure checks whether two chains, each of constant value can be perfectly discriminated by the classifier. With the new default of `split_chains=2`, perfect discrimination is no longer possible, so R* is not 2. That test can be updated to use `split_chains=1` after this PR is merged and a new version is released.

src/utils.jl
```diff
@@ -0,0 +1,96 @@
...
+"""
+function split_chain_indices(c::AbstractVector{<:Int}, split::Int=2)
+    cnew = similar(c)
+    if split == 1
+        copyto!(cnew, c)
+        return cnew
+    end
+    chains, indices = unique_indices(c)
+    chain_ind = 0
+    for (chain, inds) in zip(chains, indices)
```
devmotion, line 44: `chains`/`chain` is not used anywhere?
```suggestion
    _, indices = unique_indices(c)
    chain_ind = 0
    for inds in indices
```

src/utils.jl
```diff
@@ -0,0 +1,96 @@
...
+        # 4 items across 3 partitions because Iterators.partition(1:4, 1) == [[1], [2], [3]]
+        # and Iterators.partition(1:4, 2) == [[1, 2], [3, 4]]. But we would want
+        # [[1, 2], [3], [4]].
+        for j in 1:split
+            chain_ind += 1
+            ndraws_this_split = ndraws_per_split + (j ≤ rem)
+            i = ilast + 1
+            ilast = i + ndraws_this_split - 1
+            @views cnew[inds[i:ilast]] .= chain_ind
+        end
```
devmotion, line 57: I have the feeling it would be more efficient and easier for the compiler to just write this as a loop. That would avoid broadcasting and the views. Maybe something like the following?
```suggestion
        # here we can't use Iterators.partition because it's greedy. e.g. we can't partition
        # 4 items across 3 partitions because Iterators.partition(1:4, 1) == [[1], [2], [3]]
        # and Iterators.partition(1:4, 2) == [[1, 2], [3, 4]]. But we would want
        # [[1, 2], [3], [4]].
        i = j = 0
        ndraws_this_split = ndraws_per_split + (j < rem)
        chain_ind += 1
        for ind in inds
            cnew[ind] = chain_ind
            if (i += 1) == ndraws_this_split
                i = 0
                j += 1
                ndraws_this_split = ndraws_per_split + (j < rem)
                chain_ind += 1
            end
        end
```

src/utils.jl
```diff
@@ -0,0 +1,96 @@
...
+        split::Int=2,
+    ) -> AbstractVector{Int}
+
+Split each chain in `chain_inds` into `split` chains.
+
+For each chain in `chain_inds`, all entries are assumed to correspond to draws that have
+been ordered by iteration number. The result is a vector of the same length as `chain_inds`
+where each entry is the new index of the chain that the corresponding draw belongs to.
+"""
+function split_chain_indices(c::AbstractVector{<:Int}, split::Int=2)
```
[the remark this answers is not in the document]
sethaxen: I meant the first one. For now we constrain chain indices to be `Int` in `rstar`. If we ever relax that, we could do so here as well.

src/utils.jl
sethaxen: It's a little less readable I think, but yes, it's likely more efficient.

src/utils.jl
```diff
@@ -0,0 +1,100 @@
...
+where each entry is the new index of the chain that the corresponding draw belongs to.
+"""
+function split_chain_indices(c::AbstractVector{Int}, split::Int=2)
+    cnew = similar(c)
+    if split == 1
+        copyto!(cnew, c)
+        return cnew
+    end
+    _, indices = unique_indices(c)
+    chain_ind = 0
```
devmotion, line 43: I noticed the latest change causes an issue as it skips over some of the `chain_ind`. It could be fixed by removing `chain_ind += 1` below and setting
```suggestion
    chain_ind = 1
```

src/utils.jl
```diff
@@ -0,0 +1,100 @@
...
+    for inds in indices
+        ndraws_per_split, rem = divrem(length(inds), split)
+        # here we can't use Iterators.partition because it's greedy. e.g. we can't partition
+        # 4 items across 3 partitions because Iterators.partition(1:4, 1) == [[1], [2], [3]]
+        # and Iterators.partition(1:4, 2) == [[1, 2], [3, 4]]. But we would want
+        # [[1, 2], [3], [4]].
+        i = j = 0
+        ndraws_this_split = ndraws_per_split + (j < rem)
+        chain_ind += 1
```
devmotion, line 52:
```suggestion
```

src/utils.jl
devmotion: This can be seen by e.g. checking the output of `split_chain_indices(c)` with `c` from the tests.

src/utils.jl
devmotion: As an explanation, `chain_ind` is already updated after every split below.

src/utils.jl
devmotion: Sorry for that.

sethaxen: Ah, thanks! I added a test to cover this case.

devmotion (approved):
Looks good to me, thank you!

sethaxen:
> Looks good to me, thank you!

No problem, thanks for the detailed reviews!

sethaxen:
Ah, it seems I do not have permissions to merge, since the integration test fails. Can you merge?

PR #53, opened 2022-12-14
Improvements to ess_rhat

sethaxen:
This PR implements the nonbreaking parts of #22 (specifically, the proposal in https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/22#issuecomment-1338251834):
- makes splitting of chains optional
- support passing an `estimator` to compute an estimate-specific ESS (and R-hat too, though I don't think this is commonly used)
- Add `ess_rhat_bulk` to compute rank-normalized ESS and R-hat
- Add `ess_tail` and `rhat_tail` (the two cannot be simultaneously computed since they use different transformations)

~Currently this is marked as a draft while I write more tests and try to figure out how improve type-inferrability for utility functions.~


devmotion (commented):
I haven't checked the PR in detail yet (since I'm on my phone and tests seem to fail) but one immediate suggestion is to avoid the JLD dependency and the JLD file. Either just use a text file or even better create the data in the tests without storing it.

sethaxen:
> ...tests seem to fail

yes, the failures are in `rstar`, which is not touched here, so I'm guessing it's an issue with Random seed.

> ...one immediate suggestion is to avoid the JLD dependency and the JLD file. Either just use a text file or even better create the data in the tests without storing it.

Generating in the tests is cumbersome. When I tried sampling with Turing, I gave up after nearly an hour of sampling. StanSample can do it within a few minutes, but then we need CmdStan installed to run the tests locally and globally. Plus I wouldn't assume the tests would pass for all possible seeds to the sampler, so if CmdStan ever made a change to its sampler, our tests could suddenly begin failing. I'll add these draws as a text file.

sethaxen:
> > ...tests seem to fail
> 
> yes, the failures are in `rstar`, which is not touched here, so I'm guessing it's an issue with Random seed.

The `rstar` tests are failing on the main branch, see #55

> ...one immediate suggestion is to avoid the JLD dependency and the JLD file. Either just use a text file or even better create the data in the tests without storing it.

Do you have suggestions of how to write a multidimensional array as a text file? I usually use JLD2 or NetCDF for this.

devmotion:
Flatten it? I still think we should avoid binary files such as jld in the repo, both due to git and due to potential incompatibilities with future versions of Julia and/or JLD2 (similar to eg how JLD was abandoned and issues with BSON I ran into, even though probably they are less likely with simple arrays).

src/ess.jl
```diff
@@ -311,43 +342,18 @@ function ess_rhat(
     return ess, rhat
 end
 
-"""
-	copyto_split!(out::AbstractMatrix, x::AbstractMatrix)
-
-Copy the elements of matrix `x` to matrix `out`, in which each column of `x` is split.
+function ess_rhat_bulk(x::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
```
sethaxen, line 345: Since in the next breaking release, my goal is to split `ess_rhat` into `ess` and `rhat`, perhaps it's better to just declare `ess_bulk` and `rhat_bulk` here.

src/ess.jl
```diff
@@ -197,23 +197,54 @@ end
...
+
+## Estimators
+
+The ESS and ``\\hat{R}`` values can be computed for the following estimators:
+- `Statistics.mean`
+- `Statistics.median`
+- `Statistics.std`
+- `StatsBase.mad`
+- `Base.Fix2(Statistics.quantile, p::Real)`
+- `Base.Fix2(StatsBase.percentile, p::Real)`
```
sethaxen, line 226:
```suggestion
```

sethaxen (commented):
I found that by hand-coding the gradients I was able to get fast enough sampling to do it in the test suite. Now all tests pass except for those failing due to #56

src/ess.jl
```diff
@@ -197,23 +197,53 @@ end
...
+- `Base.Fix2(Statistics.quantile, p::Real)`
 """
-function ess_rhat(
-    chains::AbstractArray{<:Union{Missing,Real},3};
+function ess_rhat(samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    return ess_rhat(Statistics.mean, samples; kwargs...)
+end
+function ess_rhat(f, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    x = expectand_proxy(f, samples)
+    x === nothing && @error "The estimator $f is not yet supported by `ess_rhat`."
```
devmotion, line 232: IMO an `ArgumentError` would be more appropriate:
```suggestion
    if x === nothing
        throw(ArgumentError("the estimator $f is not yet supported by `ess_rhat`"))
    end
```

src/ess.jl
```diff
@@ -197,23 +197,53 @@ end
...
+    values = _ess_rhat_mean(x; kwargs...)
```
devmotion, line 233: Couldn't we just use
```suggestion
    values = ess_rhat(Statistics.mean, x; kwargs...)
```
?

src/ess.jl
```diff
@@ -197,23 +197,53 @@ end
...
+    return values
+end
+function _ess_rhat_mean(
```
devmotion, line 236:
```suggestion
function ess_rhat(
    ::typeof(Statistics.mean),
```

src/ess.jl
```diff
@@ -311,43 +341,18 @@ function ess_rhat(
...
-        ),
-    )
-    nrows_out == nrows_x ÷ 2 || throw(
-        DimensionMismatch(
-            "the output matrix must have half as many rows as as the input matrix"
-        ),
+function ess_tail(
+    x::AbstractArray{<:Union{Missing,Real},3}; tail_prob::Real=1//10, kwargs...
+)
+    T = Base.promote_eltype(x, tail_prob)
```
devmotion, line 351: What's the reason for the promotion? Shouldn't it work fine without promotion?
```suggestion
```

src/ess.jl
```diff
@@ -311,43 +341,18 @@ function ess_rhat(
...
+    return min.(
+        ess_rhat(Base.Fix2(Statistics.quantile, T(tail_prob / 2)), x; kwargs...)[1],
```
devmotion, line 353:
```suggestion
        ess_rhat(Base.Fix2(Statistics.quantile, tail_prob / 2), x; kwargs...)[1],
```

src/ess.jl
```diff
@@ -311,43 +341,18 @@ function ess_rhat(
...
+        ess_rhat(Base.Fix2(Statistics.quantile, T(1 - tail_prob / 2)), x; kwargs...)[1],
```
devmotion, line 354:
```suggestion
        ess_rhat(Base.Fix2(Statistics.quantile, 1 - tail_prob / 2), x; kwargs...)[1],
```

src/utils.jl
```diff
@@ -21,6 +21,22 @@ function unique_indices(x)
     return unique, indices
 end
 
+"""
+    split_chains(data::AbstractArray{<:Any,3}, split::Int=2)
```
devmotion, line 25: - Is it (supposed to be) exported? If not, maybe use `_split_chains` to make it clearer that it is an internal function?
- Do we need a default argument at all?

src/utils.jl
```diff
@@ -21,6 +21,22 @@ function unique_indices(x)
...
+
+Split each chain in `data` of shape `(ndraws, nchains, nparams)` into `split` chains.
+
+If `ndraws` is not divisible by `split`, the last `mod(ndraws, split)` iterations are
```
devmotion, line 29: In particular if the fumction is not exported, it would be good to explain these things in the documentation of `ess_rhat`.

src/utils.jl
```diff
@@ -21,6 +21,22 @@ function unique_indices(x)
...
+dropped. The result is a reshaped view of `data`.
+"""
+function split_chains(data::AbstractArray{<:Any,3}, split::Int=2)
+    ndraws, nchains, nparams = size(data)
+    ndraws_split, niter_drop = divrem(ndraws, split)
+    nchains_split = nchains * split
+    data_sub = @views data[firstindex(data):(end - niter_drop), :, :]
```
devmotion, line 36: Shouldn't this be the first index along the first axis only?
```suggestion
    data_sub = @views data[firstindex(data, 1):(end - niter_drop), :, :]
```
 Or on newer Julia versions simply
```suggestion
    data_sub = @views data[begin:(end - niter_drop), :, :]
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
     end
     return inds1, inds2
 end
+
+_sample_dims(data::AbstractVector) = Colon()
+_sample_dims(data::AbstractArray{<:Any,3}) = (1, 2)
+
+"""
+    fold([f,] x::AbstractVector)
```
devmotion, line 121: IMO we should use a different name. It's too similar to foldl and foldr in Base, and IMO not clear enough.

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    fold([f,] x::AbstractArray{<:Any,3}; dims=(1, 2))
+
+Compute the absolute deviation of `x` from `f(x)`, where `f` defaults to `Statistics.median`.
+
+`f` is generally a measure of central tendency. `dims` are the dimensions over which the
+estimator `f` reduces and are passed as kwargs to `f`.
+"""
+fold(f, data; dims=_sample_dims(data)) = abs.(data .- f(data; dims=dims))
```
devmotion, line 129: Dispatch on `data` since according to the docstring `dims` is not supported for `data::AbstractVector`?

Moreover, as Julia does not specialize on functions, we want to use a type parameter here, I assume:
```suggestion
function fold(f::F, data::AbstractVector) where {F}
    return abs.(data .- f(data))
end
function fold(f::F data::AbstractArray{<:Any,3}; dims=(1,2)) where {F}
    return abs.(data .- f(data; dims=dims))
end
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+fold(data; kwargs...) = fold(Statistics.median, data; kwargs...)
+
+"""
+    rank_normalize(x::AbstractVector)
```
devmotion, line 133: Again, is this (supposed to be) exported? Otherwise maybe use `_rank_normalize`?

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
     end
     return inds1, inds2
 end
+
+_sample_dims(data::AbstractVector) = Colon()
+_sample_dims(data::AbstractArray{<:Any,3}) = (1, 2)
+
+"""
+    fold([f,] x::AbstractVector)
+    fold([f,] x::AbstractArray{<:Any,3}; dims=(1, 2))
```
devmotion, line 122: Do we need a default value?

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+
+`f` is generally a measure of central tendency. `dims` are the dimensions over which the
+estimator `f` reduces and are passed as kwargs to `f`.
+"""
+fold(f, data; dims=_sample_dims(data)) = abs.(data .- f(data; dims=dims))
+fold(data; kwargs...) = fold(Statistics.median, data; kwargs...)
+
+"""
+    rank_normalize(x::AbstractVector)
+    rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2))
```
devmotion, line 134: Do we need a default value?

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+
+Rank-normalize the inputs `x` along the dimensions `dims`.
+
+Rank-normalization proceeds by first ranking the inputs using "tied ranking"
```
devmotion, line 138: Would be good to explain this in the high-level exported functions where this is used.

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+end
+function rank_normalize(x::AbstractVector)
+    values = similar(x, float(eltype(x)))
+    rank_normalize!(values, x)
+    return values
+end
+function rank_normalize!(values, x)
+    rank = StatsBase.tiedrank(x)
+    _normal_quantiles_from_ranks!(values, rank)
+    values .= StatsFuns.norminvcdf.(values)
```
devmotion, line 156: Without broadcasting?
```suggestion
    map!(StatsFuns.norminvcdf, values, values)
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2))
+
+Rank-normalize the inputs `x` along the dimensions `dims`.
+
+Rank-normalization proceeds by first ranking the inputs using "tied ranking"
+and then transforming the ranks to normal quantiles so that the result is standard
+normally distributed.
+"""
+function rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2))
+    # TODO: can we avoid mapslices and the allocations here?
```
devmotion, line 143: I guess `map` + `eachslice` is not better?

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    rank_normalize(x::AbstractVector)
+    rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2))
+
+Rank-normalize the inputs `x` along the dimensions `dims`.
+
+Rank-normalization proceeds by first ranking the inputs using "tied ranking"
+and then transforming the ranks to normal quantiles so that the result is standard
+normally distributed.
+"""
+function rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2))
```
devmotion, line 142: Maybe dispatch on `dims` (see below)?
```suggestion
rank_normalize(x::AbstractArray{<:Any,3}; dims=(1, 2)) = _rank_normalize(x, dims)
function _rank_normalize(x::AbstractArray, dims)
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    # TODO: can we avoid mapslices and the allocations here?
+    return mapslices(x; dims=dims) do xi
+        return reshape(rank_normalize(vec(xi)), size(xi))
```
devmotion, line 145: Could avoid `reshape` and `vec` if there is a generic implementation on `AbstractArray` without flattening:
```suggestion
        return _rank_normalize(xi, Colon())
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    end
+end
+function rank_normalize(x::AbstractVector)
```
devmotion, line 148:
```suggestion
rank_normalize(x::AbstractVector) = _rank_normalize(x, Colon())
function _rank_normalize(x::AbstractArray, ::Colon)
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+# "α-β correction" recommended in Eq 6.10.3 of
+# Blom. Statistical Estimates and Transformed Beta-Variables. Wiley; New York, 1958
+function _normal_quantiles_from_ranks!(q, r; α=3//8)
+    n = length(r)
+    q .= (r .- α) ./ (n - 2α + 1)
+    return q
+end
+
+"""
+    expectand_proxy(f, x::AbstractArray{<:Union{Real,Missing},3}})
```
devmotion, line 170: Same as above, is this internal? If yes, maybe rename and add explanations to high-level API.

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+function expectand_proxy(::typeof(Statistics.std), x)
+    return (x .- Statistics.mean(x; dims=(1, 2))) .^ 2
+end
+function expectand_proxy(::typeof(StatsBase.mad), x)
+    x_folded = fold(Statistics.median, x; dims=(1, 2))
+    return expectand_proxy(Statistics.median, x_folded)
+end
+function expectand_proxy(f::Base.Fix2{typeof(Statistics.quantile),<:Real}, x)
+    p = f.x
+    T = Base.promote_eltype(x, p)
```
devmotion, line 193: Is this needed? Seems it could be avoided by using `mapslices`. Or even just set to `Bool`.
```suggestion
```

src/utils.jl
```diff
@@ -97,3 +113,88 @@ function shuffle_split_stratified(
...
+    y = similar(x, T)
```
devmotion, line 194:
```suggestion
    y = similar(x, Bool)
```

src/ess.jl
sethaxen: This is a workaround for https://github.com/JuliaStats/Statistics.jl/issues/136.

src/utils.jl
sethaxen: > * Is it (supposed to be) exported?

Originally I planned for it to be, but I no longer thing that is useful.

> * If not, maybe use `_split_chains` to make it clearer that it is an internal function?
> 
>     * Do we need a default argument at all?

`split_chain_indices` similarly has no leading underscore and has a default argument, and because the methods are related, I think it makes sense to keep them consistent in this regard.

src/utils.jl
sethaxen: `begin` isn't supported on Julia v1.3 (our oldest supported version). I didn't even know `firstindex` could take a dim argument.

src/utils.jl
sethaxen: Agreed. Originally we also folded about the mean (the old approach for computing ESS of `std`), but now we only use it for `median`, so `_fold_median` may be better.

sethaxen:
@devmotion thanks for the detailed review. Since I now no longer intend for the user to directly call `fold` or `rank_normalize`, these no longer needed the `dims` keyword, which I have removed.

One other note: I removed definitions of R-hat as "potential scale reduction factor". While R-hat was originally motivated as estimating this quantity, it often fails to estimate it well but is still useful for diagnosing convergence e.g. from Vehtari et al:
image
Section 3 of https://arxiv.org/abs/2110.13017 give several other interpretations of R-hat.

Newer variants like bulk-Rhat, tail-Rhat, and their minimum are useful for diagnosing convergence but have no connection to PSRF. As a result, neither posterior nor ArviZ mention PSRF in their documentation of R-hat.

src/utils.jl
sethaxen: While BDA3 and Vehtari's paper do not discuss the case with odd chains, Stan, posterior, and ArviZ all discard the middle element. For `split_chains > 2`, we could generalize this by discarding the first `r` middle elements, where `r = rem(niter, split_chains)`. This requires bringing back `copy_split!`.

src/ess.jl
```diff
@@ -197,23 +187,67 @@ end
...
+[^VehtariGelman2021]: Vehtari, A., Gelman, A., Simpson, D., Carpenter, B., & Bürkner, P. C. (2021).
+    Rank-normalization, folding, and localization: An improved ``\\widehat {R}`` for
+    assessing convergence of MCMC. Bayesian Analysis.
+    doi: [10.1214/20-BA1221](https://doi.org/10.1214/20-BA1221)
+    arXiv: [1903.08008](https://arxiv.org/abs/1903.08008)
 """
+function ess_rhat(samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    return ess_rhat(Statistics.mean, samples; kwargs...)
+end
+function ess_rhat(f, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
```
sethaxen, line 232: A flaw of this API is that it doesn't support estimators whose behavior is modified by keyword arguments. e.g. `StatsBase.mad` can take a user-provided `center` keyword, and we could trivially update `_expectand_proxy` to handle that, but there's no corresponding `Base.Fix` wrapper that handles such keywords. Will users need that? Probably not.

src/utils.jl
```diff
@@ -21,6 +21,51 @@ function unique_indices(x)
...
+    )
+    nrows_out2, nrows_discard = divrem(nrows_x, nsplits)
+    nrows_out == nrows_out2 || throw(
+        DimensionMismatch(
+            "the output matrix must have $nsplits times as many rows as as the input matrix",
+        ),
+    )
+
+    jout = 0
+    @inbounds for j in 1:ncols_x
```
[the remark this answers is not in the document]
devmotion: Maybe `copyto!` with correct offsets etc. is safer?

src/utils.jl
```diff
@@ -97,3 +142,40 @@ function shuffle_split_stratified(
...
+
+"""
+    _fold_around_median(x::AbstractArray{<:Any,3})
+
+Compute the absolute deviation of `x` from `Statistics.median(x)`.
+"""
+_fold_around_median(data) = abs.(data .- Statistics.median(data; dims=(1, 2)))
+
+"""
+    __rank_normalizealize(x::AbstractArray{<:Any,3})
```
devmotion, line 154:
```suggestion
    _rank_normalize(x::AbstractArray{<:Any,3})
```

test/ess.jl
```diff
@@ -1,55 +1,94 @@
...
+    @testset "ESS and R̂ (IID samples)" begin
         # Repeat tests with different scales
-        for scale in (1, 50, 100)
-            x = scale * rawx
-
-            ess_standard, rhat_standard = ess_rhat(x)
-            ess_standard2, rhat_standard2 = ess_rhat(x; method=ESSMethod())
-            ess_fft, rhat_fft = ess_rhat(x; method=FFTESSMethod())
-            ess_bda, rhat_bda = ess_rhat(x; method=BDAESSMethod())
+        @testset "scale=$scale, nchains=$nchains, split_chains=$split_chains" for scale in
```
devmotion, line 63: This is the default description/name anyway IIRC.
```suggestion
        @testset for scale in
```

test/ess.jl
```diff
@@ -87,4 +126,116 @@
...
+        estimators = [mean, median, std, mad, Base.Fix2(quantile, 0.25)]
+        dists = [Normal(10, 100), Exponential(10), TDist(7) * 10 - 20]
+        # AR(1) coefficients. 0 is IID, -0.3 is slightly anticorrelated, 0.9 is highly autocorrelated
+        φs = [-0.3, -0.1, 0.1, 0.3, 0.5, 0.7, 0.9]
+        iter = filter(collect(Iterators.product(estimators, dists, φs))) do (f, dist, φ)
+            return !(f === mad) || dist isa Normal
+        end
+        nchecks = nparams * length(iter)
+        α = (0.1 / nchecks) / 2  # multiple correction
+        rng = Random.default_rng()
```
devmotion, line 161: What's the purpose of `rng`? It seems you only call `ar1` with the default RNG, so maybe just remove the `rng` argument from `ar1` and sample with `randn(T, ...)` without RNG inside the function?

test/ess.jl
```diff
@@ -87,4 +126,116 @@
...
+    end
+
+    @testset "tail- ESS and R-hat detect mismatched scales" begin
+        # simulate chains with same stationary mean but different stationary scales
+        rng = Random.default_rng()
+        φ = 0.1 # low autocorrelation
+        σs = sqrt(1 - φ^2) .* [0.1, 1, 10, 100]
+        ndraws = 1_000
+        nparams = 100
+        x = cat((ar1(rng, φ, σ, ndraws, 1, nparams) for σ in σs)...; dims=2) .+ 10
```
devmotion, line 194:
```suggestion
        x = 10 .+ mapreduce(hcat, σs) do σ
            return ar1(rng, φ, σ, ndraws, 1, nparams)
        end
```

src/ess.jl
```diff
@@ -261,10 +295,11 @@ function ess_rhat(
         W = Statistics.mean(chain_var)
 
         # compute variance estimator var₊, which accounts for between-chain variance as well
-        var₊ = correctionfactor * W + Statistics.var(chain_mean; corrected=true)
+        # avoid NaN when nchains=1
```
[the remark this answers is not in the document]
sethaxen: I noticed these as well. I'll remove `U` here. For `correctionfactor`, I'd prefer to leave it for another PR where I make sure there are no other places in our most common functions where we end up hardcoding a type like this.

src/ess.jl
```diff
@@ -312,42 +347,95 @@ function ess_rhat(
...
 
-Copy the elements of matrix `x` to matrix `out`, in which each column of `x` is split.
+For a description of `kwargs`, see [`ess_rhat`](@ref).
 
-If the number of rows in `x` is odd, the sample at index `(size(x, 1) + 1) / 2` is dropped.
+The bulk-ESS and bulk-``\\widehat{R}`` are variants of ESS and ``\\widehat{R}`` that
+diagnose poor convergence in the bulk of the distribution due to trends or different
+locations of the chains. While it is conceptually related to [`ess_rhat`](@ref) for
+`Statistics.mean`, it is well-defined even if the chains do not have finite variance.
+[^VehtariGelman2021]
```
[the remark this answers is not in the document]
sethaxen: This definitely makes sense for Markdown. For docstrings and comments, I assume the reason for wrapping is so that it's easily readable on GitHub, where wrapping doesn't occur. (Markdown is just rendered as Markdown, so not an issue there). At any rate, BlueStyle wraps docstrings at 92 characters.

src/ess.jl
```diff
@@ -312,42 +347,95 @@ function ess_rhat(
...
 
-    return out
+See also: [`ess_tail`](@ref), [`ess_rhat_bulk`](@ref)
+"""
+rhat_tail(x; kwargs...) = ess_rhat_bulk(_fold_around_median(x); kwargs...)[2]
+
+# Compute an expectand `z` such that ``\\textrm{mean-ESS}(z) ≈ \\textrm{f-ESS}(x)``.
+# If no proxy expectand for `f` is known, `nothing` is returned.
+_expectand_proxy(f, x) = nothing
+_expectand_proxy(::typeof(Statistics.mean), x) = x
```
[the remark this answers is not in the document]
sethaxen: Ah yes, this stopped being used when we dropped `_ess_mean`

src/ess.jl
```diff
@@ -7,16 +7,10 @@ abstract type AbstractESSMethod end
 The `ESSMethod` uses a standard algorithm for estimating the
 effective sample size of MCMC chains.
 
-It is is based on the discussion by Vehtari et al. and uses the
-biased estimator of the autocovariance, as discussed by Geyer.
+It is is based on the discussion by [^VehtariGelman2021] and uses the
+biased estimator of the autocovariance, as discussed by [^Geyer1992].
```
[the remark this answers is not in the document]
sethaxen: IIRC the reason I removed the duplicates was because we use `makedocs` with `strict=true`, which errors when there are duplicates. We'll need to change it to `strict=false`.

src/ess.jl
```diff
@@ -312,42 +347,95 @@ function ess_rhat(
...
+    x_folded = _fold_around_median(x)
+    return _expectand_proxy(Statistics.median, x_folded)
+end
+function _expectand_proxy(f::Base.Fix2{typeof(Statistics.quantile),<:Real}, x)
+    y = similar(x, Bool)
+    # currently quantile does not support a dims keyword argument
+    for (xi, yi) in zip(eachslice(x; dims=3), eachslice(y; dims=3))
+        yi .= xi .≤ f(vec(xi))
+    end
+    return y
```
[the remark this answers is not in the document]
devmotion: Too bad.

src/ess.jl
sethaxen: Agreed, it's inelegant, but this is something we could support in the future. I'd probably wait until a user asks for it though.

sethaxen:
I've implemented all requested changes. Only the windows tests fail, but that's expected. (see https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/56#issuecomment-1377306414)

The [references](https://turinglang.github.io/MCMCDiagnosticTools.jl/previews/PR53/) now look a little strange, but it seems to be the best we can do right now:
image

devmotion (approved):
Very nice, great PR! Looks good to me once the Linux test passes.

sethaxen:
> Very nice, great PR! Looks good to me once the Linux test passes.

Thanks for the review! Can you merge? The repo settings require all tests pass before merge, and I don't have permissions to override.

devmotion:
I changed the settings, now PRs can be merged (from everyone with merge permissions) even if Windows tests fail.

sethaxen:
Thanks!

PR #55, opened 2023-01-03
Run CI nightly

sethaxen:
CI seems to be failing on main, but this is not reflected in the CI status, since it does not run on a schedule. This PR adds a nightly run of CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - main
   pull_request:
+  schedule:
+    - cron: "0 0 * * *"
 
 jobs:
   test:


devmotion:
I would prefer closing and not merging this PR. I think generally there's no benefit from running CI on a fixed schedule, in particular if dependencies follow semver. Hence in my opinion running CI every day/week/month does not help much but rather wastes computational resources and energy. I assume for similar reasons most Julia packages are not tested on a fixed schedule.

devmotion:
IMO environmental-wise the best thing is to not consume energy if it's not needed in the first place (similar to how CO2 compensations of flights are a suboptimal workaround but the better alternative would be to not fly in the first place). I strongly believe that occasionally finding upstream bugs before they're found by users or developers is not worth it.

Another reason is that I think that it should really be the responsibility of upstream packages to ensure that non-breaking releases are actually non-breaking. In particular, it seems wrong if they rely on downstream packages for catching their bugs. I think putting more responsibility on upstream packages eventually leads to a better ecosystem since it means less work for downstream developers if packages that do not follow semver/introduce too many bugs are not used as dependencies anymore.

Another reason for not running CI on a fixed schedule is that even if tests pass every night it is not guaranteed that tests pass if users use other officially declared to be compatible versions of the dependencies.

sethaxen:
> > I assume for similar reasons most Julia packages are not tested on a fixed schedule.
> 
> I don't have data on this.

By comparing https://github.com/search?q=%22julia-runtest%22+path%3A*.yml&type=code and https://github.com/search?q=cron+%22julia-runtest%22+path%3A*.yml&type=code, I'd estimate ~5% of Julia packages run CI on a schedule.

> Another reason is that I think that it should really be the responsibility of upstream packages to ensure that non-breaking releases are actually non-breaking

As I previously noted, in my experience this is frequent, not occasional. As an example, Optimization.jl and SciMLBase.jl have made multiple breaking API changes in October that broke Pathfinder.jl, which I only know and fixed because my scheduled CI caught it. On the one hand, our devs are spread too thin and don't regularly work on every package they maintain, so they may not encounter a bug until months after it has appeared. On the other hand, many users don't open issues when they encounter bugs; they just stop using packages. If they encounter this enough, this leads some to conclude that the ecosystem is unsafe. While I agree that it's the responsibility of dependencies to follow semver, practically, mistakes will be made and often are.

Given Julia's composeability, one could argue that downstream CI failures function as integration tests, even if they shouldn't be necessary. Given how few maintainers there are working across the Julia ecosystem, automating the process of identifying these integration test failures is very useful. While I agree that we should be concerned with the consumption of our CI, I also suspect that the consumption from regularly running CI is low compared to overall consumption from running CI anytime a commit is pushed to a PR, and especially when comparing lightweight projects like this to much larger ones with more expensive CI and daily updates.

> I strongly believe that occasionally finding upstream bugs before they're found by users or developers is not worth it.

Since we strongly disagree, even on the compromise of running weekly, I'll close this PR.

PR #58, opened 2023-01-09
Improve ESS estimates for antithetic chains

### LINKED ISSUE #40: Negative ESS
In this example something seems to go wrong (check ESS parameter 4): (data at https://gist.github.com/mschauer/96ffd226d91160b1c9288200c13a77d5)

```


julia> MCMCChains.Chains(readdlm("badsamples.dat")')
Chains MCMC chain (2000×25×1 reshape(adjoint(::Matrix{Float64}), 2000, 25, 1) with eltype Float64):

Iterations        = 1:1:2000
Number of chains  = 1
Samples per chain = 2000
parameters        = param_1, param_2, param_3, param_4, param_5, param_6, param_7, param_8, param_9, param_10, param_11, param_12, param_13, param_14, param_15, param_16, param_17, param_18, param_19, param_20, param_21, param_22, param_23, param_24, param_25

Summary Statistics
  parameters      mean       std   naive_se      mcse            ess      rhat 
      Symbol   Float64   Float64    Float64   Float64        Float64   Float64 

     param_1    3.4085    1.2065     0.0270    0.0183       955.7380    0.9995
     param_2   -0.5922    0.0681     0.0015    0.0008      4861.8138    1.0003
     param_3    0.0352    0.0087     0.0002    0.0001     10913.2548    0.9995
     param_4   -0.3874    0.0859     0.0019    0.0007   -169293.0177    0.9996
     param_5    0.0045    0.0038     0.0001    0.0001      3396.2860    0.9995
     param_6   -0.2336    0.0598     0.0013    0.0005     23391.6533    0.9997
```

sethaxen:
This PR fixes #40. It's marked as draft currently because I plan to use the test utilities in #53 to write the tests.


src/ess.jl
```diff
@@ -280,10 +281,13 @@ function ess_rhat(
 
         # sum correlation estimates
         pₜ = ρ_even + ρ_odd
-        sum_pₜ = pₜ
+        sum_pₜ = zero(pₜ)
 
         k = 2
         while k < maxlag
```
sethaxen, line 287: We can rework this loop a little more to avoid computing the final `ρ_odd`, which is never used.

src/ess.jl
```diff
@@ -296,15 +300,17 @@ function ess_rhat(
...
-            sum_pₜ += pₜ
-
             # update indices
             k += 2
         end
 
+        # improved truncation reduces variance for antithetic chains
+        τ = 2 * sum_pₜ + max(0, ρ_even) - 1
+
+        # enforce ESS is in [0, ntotal*log10(ntotal)]
```
devmotion, line 310: Wouldn't it be more natural to enforce this on ESS below instead of tau?

devmotion (commented):
Thank you! A preliminary comment: could you add some references (to eg our discussion and in particular your last comment and the references therein) such that it is clear why these changes are added and where eg the bound on ESS came from?

src/ess.jl
sethaxen: It requires an additional conditional. This approach sets both negative ESS and large ESS to the maximum value of `ntotal*log10(ntotal)`. Otherwise we would use
```julia
ess_max = LogExpFunctions.xlogx(oftype(one(T), ntotal))
ess[i] = τ < 0 ? ess_max : min(ntotal / τ, ess_max) 
```

devmotion: Ah I didn't realize that negative values are not set to 0 but also the maximum value. It could be written also as
```julia
ess_i = ntotal / max(τ, zero(τ))
ess_max = ntotal * log10(oftype(ess_i, ntotal))
ess[i] = min(ess_i, ess_max)
```
(ideally with a precomputed `ess_max` of the correct type) Still requires an additional conditional but arguably looks nicer :smile:

sethaxen: Yeah, I agree this is clearer.

sethaxen:
@devmotion this is ready for final review

src/ess.jl
```diff
@@ -258,15 +260,18 @@ function ess_rhat(
...
 
-    # do not compute estimates if there is only one sample or lag
-    maxlag = min(maxlag, niter - 1)
-    if !(maxlag > 0)
+    # discard the last pair of autocorrelations, which are poorly estimated and only matter
+    # when chains have mixed poorly anyways.
+    # leave the last even autocorrelation as a bias term that reduces variance for
+    # case of antithetical chains, see below
+    maxlag = min(maxlag, niter - 4)
+    if !(maxlag > 0) || T <: Missing
```
devmotion, line 270: Isn't the only case where `T <: Missing` if `T === Missing`?
```suggestion
    if !(maxlag > 0) || T === Missing
```

src/ess.jl
```diff
@@ -328,10 +336,12 @@ function ess_rhat(
         sum_pₜ = pₜ
 
         k = 2
-        while k < maxlag
+        while true
```
devmotion, line 339: Any reason why we do not just use
```suggestion
        for k in 2:2:(maxlag - 1)
```
here?

src/ess.jl
```diff
@@ -328,10 +336,12 @@ function ess_rhat(
...
             # compute subsequent autocorrelation of all chains
             # by combining estimates of each chain
             ρ_even = 1 - inv_var₊ * (W - mean_autocov(k, esscache))
+            # stop summation if the next even lag would exceed maxlag. this ρ_odd is unused.
+            k < maxlag - 1 || break
```
devmotion, line 344: I wonder if it would be clearer to handle this last iteration separately after the `while`/`for` loop?

src/ess.jl
```diff
@@ -348,9 +358,18 @@ function ess_rhat(
...
 
+        # for antithetic chains
+        # - reduce variance by averaging truncation to odd lag and truncation to next even lag
+        # - prevent negative ESS for short chains by ensuring τ is nonnegative
+        # See discussions in:
+        # - § 3.2 of Vehtari et al. https://arxiv.org/pdf/1903.08008v5.pdf
+        # - https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/40
+        # - https://github.com/stan-dev/rstan/pull/618
+        # - https://github.com/stan-dev/stan/pull/2774
+        τ = max(0, 2 * sum_pₜ + max(0, ρ_even) - 1)
```
devmotion, line 369: Is this line "correct", i.e., desired, even if we never run the `while` loop (i.e., if `maxlag <= 2`? In these cases it would simplify to `τ = max(0, 2 * sum_pₜ)` since `ρ_even = 1`. And in contrast to the other values we would use the last `ρ_odd`.

src/ess.jl
sethaxen: Yes, if we move the final `ρ_even` computation after the loop (suggested below and now implemented), it's useful to declare `k` before the loop.

src/ess.jl
sethaxen: No it's not. I've modified to compute the final `ρ_even` only if `maxlag >= 2`. Otherwise it is set to 0 before this line.

devmotion (approved):
Thank you, looks good to me!

PR #61, opened 2023-01-10
Use biased autocov for all lags

### LINKED ISSUE #59: Using biased autocov for all lags
Currently we use the estimate of autocov with denominator `n-1`, which is unbiased for lag `k = 0` and biased for `k > 1`:
https://github.com/TuringLang/MCMCDiagnosticTools.jl/blob/57cacc1f54b0557d8019a312e2bd83003f9d8f01/src/ess.jl#L10-L13

As noted, Geyer and Vehtari suggest using `n` as the denominator, not `n-1`. This is also used in Stan's, posterior's, and ArviZ's implementations of ESS. Unless we have a reference with simulations for our choice choice, I propose we instead use `n-1` as suggested by these sources.

With #58 and #53, this is the last change needed to make our estimates identical to posterior's and ArviZ's within floating point precision.

sethaxen:
Fixes #59 for `ESSMethod` and `FFTESSMethod`. Since `BDAESSMethod` uses the unbiased estimator for all lags ~(I guess for historical purposes)~ as specified in BDA3, I have left it untouched.


test/ess.jl
```diff
@@ -9,6 +9,18 @@ using Statistics
 using StatsBase
 using Test
 
+struct ExplicitESSMethod <: MCMCDiagnosticTools.AbstractESSMethod end
```
devmotion, line 12: Nice!

devmotion (approved):

PR #62, opened 2023-01-12
Fixes to return types of ess_rhat

sethaxen:
This PR makes 2 improvements to `ess_rhat`:
- Don't unnecessarily promote the return type to `Float64`. 
- Remove unused type parameter `U`
- Return arrays `ess` and `rhat` preserve structure of last dimension of input array. This is especially useful when the input array is one with named dimensions, like `DimensionalData.DimArray`, since the indices in that dimension will be preserved.


src/ess.jl
```diff
@@ -255,37 +255,39 @@ function ess_rhat(
...
-    correctionfactor = (niter - 1) / niter
+    correctionfactor = (niter - 1)//niter
 
     # define cache for the computation of the autocorrelation
     esscache = build_cache(method, samples, chain_var)
 
     # define output arrays
-    ess = Vector{T}(undef, nparams)
-    rhat = Vector{T}(undef, nparams)
+    ess = similar(params, T)
```
devmotion, line 281:
```suggestion
    ess = similar(chains, T, (axes(chains, 3),))
```

src/ess.jl
```diff
@@ -255,37 +255,39 @@ function ess_rhat(
...
+    rhat = similar(ess)
+    i0 = firstindex(ess) - 1
 
     # for each parameter
     for (i, chains_slice) in enumerate(eachslice(chains; dims=3))
         # check that no values are missing
         if any(x -> x === missing, chains_slice)
-            rhat[i] = missing
-            ess[i] = missing
+            rhat[i0 + i] = missing
```
devmotion, line 289: Does that line work at all? Can `rhat` store `missing`?

src/ess.jl
```diff
@@ -255,37 +255,39 @@ function ess_rhat(
...
 
     # define cache for the computation of the autocorrelation
     esscache = build_cache(method, samples, chain_var)
 
     # define output arrays
-    ess = Vector{T}(undef, nparams)
-    rhat = Vector{T}(undef, nparams)
+    ess = similar(params, T)
+    rhat = similar(ess)
+    i0 = firstindex(ess) - 1
```
devmotion, line 283: Is `i` used anywhere in the loop below? If not it would be simpler (I think) to replace
```julia
for (i, chains_slice) in enumerate(eachslice(chains; dims=3))
```
with
```julia
for (i, chains_slice) in zip(eachindex(ess), eachslice(chains; dims=3))
```

src/ess.jl
sethaxen: Ah, hm, not it seems this is a line currently not tested. This check should probably be moved to outside the loop.

src/ess.jl
```diff
@@ -255,37 +255,39 @@ function ess_rhat(
 )
     # compute size of matrices (each chain may be split!)
     niter = size(chains, 1) ÷ split_chains
-    nparams = size(chains, 3)
     nchains = split_chains * size(chains, 2)
     ntotal = niter * nchains
 
+    # vector used to construct output vector types
+    params = view(chains, firstindex(chains, 1), firstindex(chains, 2), :)
+
```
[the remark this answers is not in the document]
sethaxen: How do you expect the `view` could be problematic? Since we're using `eachslice`, we're already using views of this array.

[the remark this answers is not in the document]
devmotion: But is that the case just because DimensionalData and AxisArrays do not implement `similar` correctly (or, rather, not all methods that would be needed)? According to the docstring of `similar` it seems we want to do what this method was designed for:
```julia
 similar(storagetype, axes)

  Create an uninitialized mutable array analogous to that specified by storagetype, but with axes specified by the last argument.

  Examples:

  similar(Array{Int}, axes(A))

  creates an array that "acts like" an Array{Int} (and might indeed be backed by one), but which is indexed identically to A. If A has conventional indexing, this will be identical to Array{Int}(undef,
  size(A)), but if A has unconventional indexing then the indices of the result will match A.

  similar(BitArray, (axes(A, 2),))

  would create a 1-dimensional logical array whose indices match those of the columns of A.
```
That's also how OffsetArratys implements it:
```julia
julia> using OffsetArrays

julia> x = OffsetArray(randn(3, 5), -1:1, 2:6)
3×5 OffsetArray(::Matrix{Float64}, -1:1, 2:6) with eltype Float64 with indices -1:1×2:6:
  0.143167   0.0881142  -1.24245   -0.190666   0.021432
 -0.694397   1.32132    -0.447271  -0.196665   0.46701
  1.59353   -0.0241462   1.54666   -2.07788   -0.403454

julia> similar(x, Missing, (axes(x, 2),))
5-element OffsetArray(::Vector{Missing}, 2:6) with eltype Missing with indices 2:6:
 missing
 missing
 missing
 missing
 missing

julia> axes(similar(x, Missing, (axes(x, 2),)))
(OffsetArrays.IdOffsetRange(values=2:6, indices=2:6),)
```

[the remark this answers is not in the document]
devmotion: There seems to be an additional (funny) behaviour of AxisArrays:
```julia
julia> using AxisArrays

julia> aa = AxisArray(randn(10, 4, 2));

julia> AxisArrays.axes(aa, 3)
Axis{:page, Base.OneTo{Int64}}(Base.OneTo(2))

julia> similar(aa, Int, (AxisArrays.axes(aa, 3),))
3-dimensional AxisArray{Int64,3,...} with axes:
    :row, Base.OneTo(10)
    :col, Base.OneTo(4)
    :page, Base.OneTo(2)
And data, a 10×4×2 Array{Int64, 3}:
[:, :, 1] =
 140405193817504  140405014786992  140405014786992  140405014786992
               0                1                1                1
 140405014758288  140405014786992  140405014786992  140405014786992
               0                1                1                1
 140405201715984  140405014786992  140405014786992  140405014786992
               0                1                1                1
 140405014786992  140405014786992  140405014786992  140405014786992
               1                1                1                1
 140405014786992  140405014786992  140405014786992  140405014786992
               1                1                1                1

[:, :, 2] =
 140405014786992  140405014786992  140405014786992  140405014786992
               1                1                1                1
 140405014786992  140405014786992  140405014786992  140405014786992
               1                1                1                1
 140405014786992  140405014786992  140405014786992  140405014786992
               1                1                1                1
 140405014786992  140405015840704  140405014786992  140405014786992
               1                0                1                1
 140405014786992  140405201715984  140405014786992                0
               1                0                1                0
```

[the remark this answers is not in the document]
sethaxen: Okay, I see what you mean now, and I agree that this should be supportable by DimensionalData (see https://github.com/rafaqz/DimensionalData.jl/issues/437). I'll make the changes here.

src/ess.jl
sethaxen: Ah yes, this line does work, because this conditional can only evaluate to `true` if `Missing <: eltype(chains)`, in which case `Missing <: T`.

src/ess.jl
```diff
@@ -255,33 +255,34 @@ function ess_rhat(
...
-    nparams = size(chains, 3)
     nchains = split_chains * size(chains, 2)
     ntotal = niter * nchains
+    axes_out = (axes(chains, 3),)
 
     # do not compute estimates if there is only one sample or lag
     maxlag = min(maxlag, niter - 1)
-    maxlag > 0 || return fill(missing, nparams), fill(missing, nparams)
+    if !(maxlag > 0)
+        return similar(chains, Missing, axes_out), similar(chains, Missing, axes_out)
```
devmotion, line 265: Do we have any tests for this? And any `@test_inferred` checks - it seems they should fail due to this line?

devmotion (commented):
Looks good, just one minor question. Possibly something for a different PR though.

src/ess.jl
sethaxen: Yes, these lines are tested: https://github.com/TuringLang/MCMCDiagnosticTools.jl/blob/b6e8de9a6ac330c4f7f6cc45185d1e509127fa1f/test/ess.jl#L122-L134.

I added more tests that the array type returned here preserves the final axis and some `@inferred` tests specifying the type union.

devmotion (approved):

src/ess.jl
sethaxen: I agree, this is a good idea.

PR #63, opened 2023-01-16
Redesign of MCSE

sethaxen:
Following https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/39#issuecomment-1380537474 and #53, this PR redesigns `mcse`:
- `mcse` now requires the user provide an estimator to make it less ambiguous what is estimated.
- the vector methods to `mcse` have been removed and replaced with 3d array methods similar to `ess_rhat`, which also allow for eltypes to include `Missing`. This allows MCSEs to be computed for many parameters with one function call and also allows for improved ESS estimates when multiple chains are provided
- a subsampling bootstrap method has been added as a fallback, since unlike batch means or overlapping batch means, it works for any estimator.
- custom estimators have been added for `mean`, `std`, and `quantile` using the ESS methods
- batch-means, initial monotone sequence, and initial positive sequence methods have been removed. These estimators only can estimate the MCSE of `mean`. From the Geyer reference in the docstrings, IMSE is better than IPSE, and both are better than batch-means, which tends to underestimate MCSE more severely. IMSE is used in the `ess_rhat(::typeof(mean), x)` call used in the new `ess_rhat` implementations.

Marked as a draft for now until I add tests.


src/mcse.jl
```diff
@@ -1,72 +1,121 @@
...
+"""
+function mcse_sbm(
+    f,
+    x::AbstractArray{<:Union{Missing,Real},3};
+    batch_size::Int=floor(Int, sqrt(size(x, 1) * size(x, 2))),
+)
+    T = promote_type(eltype(x), typeof(zero(eltype(x)) / 1))
+    values = similar(x, T, (axes(x, 3),))
+    for (i, xi) in zip(eachindex(values), eachslice(x; dims=3))
+        values[i] = _mcse_sbm(f, vec(xi); batch_size=batch_size)
```
sethaxen, line 109: I haven't seen in the literature a study of how best to combine MCMC chains when estimating MCSE with SBM. I benchmarked a few alternatives and found this one underestimated the MCSE the least: https://github.com/arviz-devs/arviz/pull/1974#discussion_r1071645315

test/mcse.jl
```diff
@@ -1,25 +1,41 @@
...
+using Distributions
+using MCMCDiagnosticTools
+using Statistics
+using StatsBase
 
-    @testset "warning" begin
-        for size in (51, 75, 100, 153)
-            @test_logs (:warn,) mcse(samples; method=:bm, size=size)
+@testset "mcse.jl" begin
+    @testset "estimand is within interval defined by MCSE estimate" begin
```
sethaxen, line 8: For `mean`, `std`, and `quantile`, these tests are somewhat redundant with those for `ess_rhat`, since only difference is that for `ess_rhat` we compute the exact asymptotic variance of the estimator since we know the stationary distribution, whereas for `mcse` we estimate the asymptotic variance of the estimator. If we want to speed up our test suite, it may be better to remove the corresponding `ess_rhat` tests and use these as integration tests.

src/mcse.jl
```diff
@@ -1,72 +1,127 @@
...
-        Ghat > 0 || break
-        value += 2 * Ghat
-    end
+`batch_size` indicates the size of the overlapping batches used to estimate the MCSE,
+defaulting to `floor(Int, sqrt(draws * chains))`.
 
-    mcse = sqrt(value / n)
+!!! note
+    SBM tends to underestimate the MCSE, especially for highly autocorrelated chains.
+    SBM should only be used as a fallbeck when a specific [`mcse`](@ref) method for
```
sethaxen, line 96: This begs the question whether we should include `mcse_sbm` at all. On the one hand it's useful to have a fallback, and I don't know of any other general-purpose methods for estimating MCSE. On the other hand, we would prefer a fallback that errs on the side of overestimating MCSE instead of overestimating.

I lean towards including it with this caveat clearly documented.

sethaxen:
There are a few cases that still need tests, but this is ready for an initial review.

src/gewekediag.jl
```diff
@@ -22,9 +22,12 @@ function gewekediag(x::AbstractVector{<:Real}; first::Real=0.1, last::Real=0.5,
     n = length(x)
     x1 = x[1:round(Int, first * n)]
     x2 = x[round(Int, n - last * n + 1):n]
-    z =
-        (Statistics.mean(x1) - Statistics.mean(x2)) /
-        hypot(mcse(x1; kwargs...), mcse(x2; kwargs...))
+    T = float(eltype(x))
+    s = hypot(
+        Base.first(mcse(Statistics.mean, reshape(x1, :, 1, 1); split_chains=1, kwargs...)),
+        Base.first(mcse(Statistics.mean, reshape(x2, :, 1, 1); split_chains=1, kwargs...)),
```
devmotion, line 28:
```suggestion
        Base.first(mcse(Statistics.mean, reshape(x1, Val(3)); split_chains=1, kwargs...)),
        Base.first(mcse(Statistics.mean, reshape(x2, Val(3)); split_chains=1, kwargs...)),
```
(the `Base` was a bit surprising before I noticed the variable named `first`)

src/gewekediag.jl
```diff
@@ -22,9 +22,12 @@ function gewekediag(x::AbstractVector{<:Real}; first::Real=0.1, last::Real=0.5,
...
+    )::T
```
devmotion, line 29: Can you explain why this is needed?

src/heideldiag.jl
```diff
@@ -17,7 +17,9 @@ function heideldiag(
     n = length(x)
     delta = trunc(Int, 0.10 * n)
     y = x[trunc(Int, n / 2):end]
-    S0 = length(y) * mcse(y; kwargs...)^2
+    T = float(eltype(x))
+    s = first(mcse(Statistics.mean, reshape(y, :, 1, 1); split_chains=1, kwargs...))::T
```
devmotion, line 21:
```suggestion
    s = first(mcse(Statistics.mean, reshape(y, Val(3)); split_chains=1, kwargs...))::T
```

src/heideldiag.jl
```diff
@@ -33,7 +35,8 @@ function heideldiag(
         end
         i += delta
     end
-    halfwidth = sqrt(2) * SpecialFunctions.erfcinv(alpha) * mcse(y; kwargs...)
+    s = first(mcse(Statistics.mean, reshape(y, :, 1, 1); split_chains=1, kwargs...))::T
```
devmotion, line 38:
```suggestion
    s = first(mcse(Statistics.mean, reshape(y, Val(3)); split_chains=1, kwargs...))::T
```

src/mcse.jl
```diff
@@ -1,72 +1,127 @@
+Base.@irrational normcdf1 0.8413447460685429486 StatsFuns.normcdf(big(1))
+Base.@irrational normcdfn1 0.1586552539314570514 StatsFuns.normcdf(big(-1))
```
devmotion, line 2: Defining `@irrational`s outside of base is actually type piracy - it will add implementations for `Irrational{:normcdf1}` etc. which is not owned by us. We decided to do it anyways in IrrationalConstants but core developers told us `@irrational`/`Irrational` was not supposed to be used outside of base. Hence if possible I'd like to avoid these definitions here:
```suggestion
```

devmotion (commented):
A very quick initial review.

src/gewekediag.jl
sethaxen: Just because of the type-inferrability coming from low `maxlag` maybe returning `missing`, which wouldn't be supported by the method (it currently always returns a `Real`). But this is unnecessary if we first merge a PR throwing an error instead. I'll open one.

src/gewekediag.jl
sethaxen: Is the `reshape(A, ::Val)` method documented anywhere? I hadn't seen it before and couldn't find it in the Julia docs. I also found this old discussion where the consensus seemed to be to replace it: https://github.com/JuliaLang/julia/issues/27474

src/mcse.jl
sethaxen: Ah, okay. In this case we round to an `Int` eventually anyways, so it's fine to use a `Float64`.
```suggestion
const normcdf1 = 0.841344746068543  # StatsFuns.normcdf(1)
const normcdfn1 = 0.15865525393145702  # StatsFuns.normcdf(-1)
```

src/gewekediag.jl
devmotion: I used it several times (was quite convenient sometimes) but I didn't know that there was some discussion about removing it. Maybe let's keep `reshape(x1, :, 1, 1)` in that case.

test/mcse.jl
sethaxen: I've opted to speed up our test suite by decreasing the number of replicates (params) for each suite to 100 instead of 1,000.

src/ess.jl
```diff
@@ -480,7 +482,7 @@ function _expectand_proxy(::typeof(StatsBase.mad), x)
     return _expectand_proxy(Statistics.median, x_folded)
 end
 function _expectand_proxy(f::Base.Fix2{typeof(Statistics.quantile),<:Real}, x)
-    y = similar(x, Bool)
+    y = similar(x)
```
sethaxen, line 485: Necessary because otherwise our transformation discards the type of `x`, so a `Float32` eltype `x` will get a `Float64` ESS.

src/mcse.jl
devmotion: Since we only want it to be used as a fallback, maybe we do not even want to add another function for it but also call it `mcse`? And just document there that it is used as a fallback but one should be aware of its limitations.

src/mcse.jl
```diff
@@ -1,72 +1,142 @@
...
+    batch_size::Int=floor(Int, sqrt(size(x, 1) * size(x, 2))),
+)
+    T = promote_type(eltype(x), typeof(zero(eltype(x)) / 1))
+    values = similar(x, T, (axes(x, 3),))
+    for (i, xi) in zip(eachindex(values), eachslice(x; dims=3))
+        values[i] = _mcse_sbm(f, vec(xi); batch_size=batch_size)
+    end
+    return values
+end
+function _mcse_sbm(f, x; batch_size)
```
devmotion, line 133: I guess we might want to specialize on `f` here?
```suggestion
function _mcse_sbm(f::F, x; batch_size) where {F}
```

src/utils.jl
```diff
@@ -145,7 +145,13 @@ end
 
 Compute the absolute deviation of `x` from `Statistics.median(x)`.
 """
-_fold_around_median(data) = abs.(data .- Statistics.median(data; dims=(1, 2)))
+function _fold_around_median(x)
+    y = similar(x)
```
devmotion, line 149: I guess for the same reasons as above? Maybe copy the comment?

src/mcse.jl
sethaxen: Good idea

src/mcse.jl
sethaxen: Probably only if we see a performance benefit, right? Otherwise it seems not specializing may be better for the compiler. https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

sethaxen: On realistic chains, specializing on `f` seems to decrease performance:
```julia
julia> x = randn(1000, 4, 100);

julia> @btime $(MCMCDiagnosticTools._mcse_sbm)($mean, $x);  # without specialization
  9.160 ms (201 allocations: 8.69 KiB)

julia> @btime $(MCMCDiagnosticTools._mcse_sbm)($mean, $x);  # with specialization
  9.632 ms (201 allocations: 8.69 KiB)
```

devmotion: Yes, seems better. Was just working on some other project and there it was useful to enforce specialization, so I was wondering about this method :slightly_smiling_face:

devmotion (approved):

PR #64, opened 2023-01-18
Raise an error if maxlag too low

sethaxen:
Instead of returning a vector of `missing` if the user provides too low of a `maxlag` to `ess_rhat`, which annoyingly ruins type-inferrability, this PR raises an error.


src/ess.jl
```diff
@@ -266,10 +268,17 @@ function ess_rhat(
...
     # leave the last even autocorrelation as a bias term that reduces variance for
     # case of antithetical chains, see below
-    maxlag = min(maxlag, niter - 4)
-    if !(maxlag > 0) || T === Missing
-        return similar(chains, Missing, axes_out), similar(chains, Missing, axes_out)
+    if !(niter > 4)
+        throw(ArgumentError("number of draws after splitting must >4 but is $niter."))
     end
+    maxlag > 0 || throw(DomainError(maxlag, "maxlag must be >0."))
+    maxlag = min(maxlag, niter - 4)
```
devmotion, line 275: Maybe move the checks to the top of the function?

devmotion (approved):
Looks good to me, just one final suggestion which you may ignore if you don't agree :slightly_smiling_face:

src/ess.jl
sethaxen: I like keeping these checks together and using `niter`, which is computed with the other sizes, so I prefer to leave here.

PR #67, opened 2023-01-23
Update TagBot workflow

sethaxen:
This PR updates the TagBot workflow to the latest recommended file (see https://github.com/JuliaRegistries/TagBot#setup).

The main reason for this is the added ability to specify the lookback when manually triggering TagBot (TagBot has had issues recently, so some of our more recent registered versions have not had tagged releases, and with this change we can trigger TagBot to generate those older releases)

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -5,7 +5,22 @@ on:
     types:
       - created
   workflow_dispatch:
-
+    inputs:
+      lookback:
+        default: 3
+permissions:
+  actions: read
+  checks: read
+  contents: write
+  deployments: read
+  issues: read
+  discussions: read
+  packages: read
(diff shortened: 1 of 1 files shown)


devmotion (approved):
Sure, let's try it. I'm not too optimistic though, I've seen issues (mainly https://github.com/JuliaRegistries/TagBot/issues/242, I think) recently also in a few packages that used the recommended setup.

sethaxen:
> I'm not too optimistic though, I've seen issues (mainly [JuliaRegistries/TagBot#242](https://github.com/JuliaRegistries/TagBot/issues/242), I think) recently also in a few packages that used the recommended setup.

I just saw one work and thought maybe it was working again, but it seems not. Either way, having this is useful for when TagBot does start working again.

PR #70, opened 2023-02-03
Do not include MCMCDiagnosticTools in test and doc environment

devmotion:
It's not necessary to include MCMCDiagnosticTools in docs/Project.toml and test/Project.toml. In the first case, we check out the package at ".." anyway (so fixing it to a version is not very useful) and in the latter Pkg adds the package automatically. On the other hand, adding it to these environments is annoying since CompatHelper will open PRs if the compat statement includes only unreleased versions.

Fixes #68. Fixes #69. Fixes #66. Fixes #65.


sethaxen (approved):
LGTM, assuming tests pass

devmotion:
Julia nightly takes forever, I already restarted it once. I don't see how that could be related to this PR though, so I suspect the same problem will show up on the main branch.

devmotion:
Maybe it's the same problem as in https://github.com/JuliaLang/julia/issues/48488.

Issue #23, opened 2021-10-27
Nested R-hat


sethaxen:
Nested R-hat is a variant of R-hat used when sampling many short chains: https://arxiv.org/abs/2110.13017

PR #72, opened 2023-02-23
Use keyword syntax for ess, ess_rhat, rhat, and mcse

sethaxen:
This PR implements the new keyword-based API proposed in https://github.com/TuringLang/MCMCDiagnosticTools.jl/issues/22#issuecomment-1398055191.


src/ess.jl
```diff
@@ -195,19 +195,21 @@ function mean_autocov(k::Int, cache::BDAESSCache)
 end
 
 """
-    ess_rhat(
-        [estimator,]
+    ess(
```
sethaxen, line 198: This seems to be a case where type inference does not do well. The type of `ess(randn(100, 10, 4); type=:tail)` is inferred to be `Any` even though every method that can be called would all return `Vector{Float64}`. If we use `Base.@constprop :agressive` or `@inline` here, then `foo(x) = ess(x; type=:tail)` becomes type-inferrable, but still not the version with the keyword argument.

sethaxen: I guess one alternative is to define the output objects in the outer methods and then the internal method are all mutating `_exp!`. But it feels like there should be a way to make this type-inferrable.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.3.0-DEV"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
devmotion, line 8: Maybe it's clear once I read through the changes below, but why do we need Compat?

src/MCMCDiagnosticTools.jl
```diff
@@ -16,18 +17,19 @@ using Statistics: Statistics
...
 export bfmi
 export discretediag
-export ess_rhat, ess_rhat_bulk, ess_tail, rhat_tail, ESSMethod, FFTESSMethod, BDAESSMethod
+export ess, ess_rhat, ESSMethod, FFTESSMethod, BDAESSMethod
 export gelmandiag, gelmandiag_multivariate
 export gewekediag
 export heideldiag
 export mcse
 export rafterydiag
+export rhat
```
devmotion, line 26: Maybe this export could be moved to `ess` and `ess_rhat` above, since they seem to belong together?

src/ess_rhat.jl
```diff
@@ -195,25 +200,23 @@ function mean_autocov(k::Int, cache::BDAESSCache)
...
     )
 
-Estimate the effective sample size and ``\\widehat{R}`` of the `samples` of shape
+Estimate the effective sample size (ESS) of the `samples` of shape
 `(draws, chains, parameters)` with the `method`.
 
-By default, the computed ESS and ``\\widehat{R}`` values correspond to the estimator `mean`.
-Other estimators can be specified by passing a function `estimator` (see below).
+Optionally, only one of the `type` of ESS estimate to return or the `estimator` for which
+ESS is computed can be specified (see below). Some `type`s accept additional `kwargs`.
```
devmotion, line 217: I was a bit confused by the first sentence. Maybe it could be reformulated a bit, maybe something like
```suggestion
Optionally, one can specify either the `type` of the ESS estimate or the `estimator` for which
ESS is computed (see below). Some `type`s accept additional `kwargs`.
```

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+    convergence in the bulk of the distribution due to trends or different locations of the
+    chains.
+- `:tail`: minimum of the quantile-ESS for the symmetric quantiles where
+    `tail_prob=0.1` is the probability in the tails. This type diagnoses poor convergence in
+    the tails of the distribution. If this type is chosen, `kwargs` may contain a
+    `tail_prob` keyword.
+- `:basic`: basic ESS, equivalent to specifying `estimator=Statistics.mean`.
+
+While Bulk-ESS is conceptually related to basic ESS, it is well-defined even if the chains
+do not have finite variance.[^VehtariGelman2021]. For each parameter, rank-normalization
```
devmotion, line 252:
```suggestion
do not have finite variance.[^VehtariGelman2021] For each parameter, rank-normalization
```

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+
 [^VehtariGelman2021]: Vehtari, A., Gelman, A., Simpson, D., Carpenter, B., & Bürkner, P. C. (2021).
     Rank-normalization, folding, and localization: An improved ``\\widehat {R}`` for
     assessing convergence of MCMC. Bayesian Analysis.
     doi: [10.1214/20-BA1221](https://doi.org/10.1214/20-BA1221)
     arXiv: [1903.08008](https://arxiv.org/abs/1903.08008)
 """
-function ess_rhat(samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
-    return ess_rhat(Statistics.mean, samples; kwargs...)
+@constprop :aggressive function ess(
```
devmotion, line 263: Why is `@constprop` needed?

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+    samples::AbstractArray{<:Union{Missing,Real},3};
+    estimator=nothing,
+    type=nothing,
```
devmotion, line 266: Maybe restrict the types of the keyword arguments here since we know the admissible types? Also, maybe it would be simpler to use `:basic` as the default value here?
```suggestion
    type::Symbol = :basic,
```
I'm not a fan of supporting both `Val` and `Symbol` but in that case we could use
```suggestion
    type::Union{Val,Symbol}=Val(:basic),
```

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+end
+function _ess(::Val{T}, ::AbstractArray{<:Union{Missing,Real},3}; kwargs...) where {T}
+    return throw(ArgumentError("the `type` `$T` is not supported by `ess`"))
+end
+function _ess(type::Val{:basic}, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    return first(_ess_rhat(type, samples; kwargs...))
+end
+function _ess(type::Val{:bulk}, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    return first(_ess_rhat(type, samples; kwargs...))
+end
```
devmotion, line 294: Couldn't we just use
```suggestion
function _ess(type::Val, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
    return first(_ess_rhat(type, samples; kwargs...))
end
```
and error if `_ess_rhat` does not support the `type`?

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
 - `StatsBase.mad`
 - `Base.Fix2(Statistics.quantile, p::Real)`
 
+## Types
+
+If no `estimator` is provided, the following `type`s of ESS estimates may be computed:
+- `:bulk`/`:rank`: mean-ESS computed on rank-normalized draws. This type diagnoses poor
```
devmotion, line 242: What's the reason for supporting both `:bulk` and `:rank`? Why not just one of them?

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+)
+    if estimator !== nothing && type !== nothing
+        throw(ArgumentError("only one of `estimator` and `type` can be specified"))
+    elseif estimator !== nothing
+        return _ess(estimator, samples; kwargs...)
+    elseif type !== nothing
+        return _ess(_val(type), samples; kwargs...)
+    else
+        return _ess(Val(:basic), samples; kwargs...)
+    end
```
devmotion, line 277: Could we simplify the logic by just saying, if `estimator !== nothing` it is used and otherwise the `type` determines what ESS estimate is returned?
```suggestion
    if estimator !== nothing
        return _ess(estimator, samples; kwargs...)
    else
        return _ess(_val(type), samples; kwargs...)
    end
```
(Again, noting that I'm not a big fan of supporting both `Val` and `Symbol`)

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+end
+function _ess(::Val{:rank}, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    return _ess(Val(:bulk), samples; kwargs...)
+end
+
+"""
+    rhat(samples::AbstractArray{Union{Real,Missing},3}; type=:rank, split_chains=2)
+
+Compute the ``\\widehat{R}`` diagnostics for each parameter in `samples` of shape
+`(chains, draws, parameters)`. [^VehtariGelman2021]
```
devmotion, line 317:
```suggestion
`(chains, draws, parameters)`.[^VehtariGelman2021]
```

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+    due to different scales of the chains.
+- `:basic`: Classic ``\\widehat{R}``.
+
+[^VehtariGelman2021]: Vehtari, A., Gelman, A., Simpson, D., Carpenter, B., & Bürkner, P. C. (2021).
+    Rank-normalization, folding, and localization: An improved ``\\widehat {R}`` for
+    assessing convergence of MCMC. Bayesian Analysis.
+    doi: [10.1214/20-BA1221](https://doi.org/10.1214/20-BA1221)
+    arXiv: [1903.08008](https://arxiv.org/abs/1903.08008)
+"""
+@constprop :aggressive function rhat(
```
devmotion, line 343: Why is `@constprop` needed here?

src/ess_rhat.jl
```diff
@@ -233,25 +236,206 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+    samples::AbstractArray{<:Union{Missing,Real},3}; type=Val(:rank), kwargs...
```
devmotion, line 344: Again could we just support `Symbol`? Or only `Val`? In any case, I think we could add a type restriction to `type`.

src/utils.jl
```diff
@@ -185,3 +190,6 @@ function _normal_quantiles_from_ranks!(q, r; α=3//8)
     q .= (r .- α) ./ (n - 2α + 1)
     return q
 end
+
+_val(k) = Val(k)
```
devmotion, line 194: Same concerns as above about supporting both `Symbol` and `Val`. I guess it would be reasonable though to constrain it to
```suggestion
_val(k::Symbol) = Val(k)
```

Project.toml
sethaxen: For `@constprop`, which was added in v1.7.

src/ess_rhat.jl
sethaxen: If we don't have it, then functions like `foo(x) = ess(x; type=:bulk)` are not type-inferrable.

src/ess_rhat.jl
sethaxen: To simplify both usage and testing of `ess`, `ess_rhat`, and `rhat`. `ess_rhat` and `rhat` both support `type=:rank` and `type=:bulk`. In both cases bulk-ESS is returned, while the R-hat value changes. It makes sense for `ess` to treat `:rank` as a synonym for `:bulk`.

src/ess_rhat.jl
sethaxen: I do think it's important to throw an error if the user tries to specify both `type` and `estimator`. A bulk-median-ESS is meaningless.

src/MCMCDiagnosticTools.jl
```diff
@@ -16,18 +17,19 @@ using Statistics: Statistics
 
 export bfmi
 export discretediag
-export ess_rhat, ess_rhat_bulk, ess_tail, rhat_tail, ESSMethod, FFTESSMethod, BDAESSMethod
+export ess, ess_rhat, ESSMethod, FFTESSMethod, BDAESSMethod
```
sethaxen, line 20:
```suggestion
export ess, ess_rhat, rhat, ESSMethod, FFTESSMethod, BDAESSMethod
```

src/ess_rhat.jl
devmotion: Wouldn't it make more sense to only support `:bulk` for `ess`? IMO it is a bit strange to support `:rank` just because `rhat` does and `ess` and `rhat` are somewhat connected via `ess_rhat` which only exists to reduce redundant computations and improve performance.

IMO it would be natural to only support `:bulk` when calling `ess`. And to support `:rank` as well in `ess_rhat` and `rhat`, and document (and implement it such) that in this case `ess` with `:bulk` is returned.

src/ess_rhat.jl
devmotion: Sure, but do you know why? Shouldn't this be handled by constant propagation? Is this a bug/deficiency of the Julia compiler that should be fixed? Or do we have some code pattern that prevents constant propagation and could be fixed to help the compiler?

sethaxen: I don't fully understand why default const-propagation doesn't work here. I'd asked about it on Julia Slack (https://julialang.slack.com/archives/C6A044SQH/p1677318248072949), and this was the recommended solution. An alternative would be to use ValSplit.jl, but that seemed to be a more complex solution.

src/ess_rhat.jl
devmotion: Hmm I was thinking that it would be sufficient to document that `estimator` would have higher precedence than `type` but I can see your point.

That makes me wonder though: Do we actually need two keyword arguments? It seems one would be sufficient if one must not specify both anyway. We could just interpret values of type `Symbol` (or `Val`) as the `type` and everything else as `estimator`?

sethaxen: I liked the explicitness of calling it an estimator, especially for consistency with `mcse`. If we use `type` for `ess` as well, then we should also use `type` for `mcse`.

src/ess_rhat.jl
sethaxen: I don't feel strongly about this. I'll drop `:rank` from `ess`.

src/ess_rhat.jl
devmotion: Type inference seems to work if you branch explicitly instead of using `Val`: https://julialang.slack.com/archives/C6A044SQH/p1677365245386389?thread_ts=1677318248.072949&cid=C6A044SQH

src/ess_rhat.jl
devmotion: Maybe that would be a good approach? An alternative could be to call it `method`.

src/ess_rhat.jl
sethaxen: I adopted this approach.

src/ess_rhat.jl
sethaxen: I prefer not to use `method`, as to me `method` is more about how an estimate is made than about what the estimate is (e.g. `method` currently used for autocovariance estimate method). I merged `estimator` and `type` as a new keyword `kind`. `kind` is about as semantically overloaded as `type` but avoids being potentially misunderstood as the type of the return object (also doesn't get colored strangely in syntax highlighters).

src/ess_rhat.jl
```diff
@@ -239,19 +257,184 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+        return _rhat(Val(:basic), samples; kwargs...)
+    else
+        return throw(ArgumentError("the `kind` `$kind` is not supported by `rhat`"))
+    end
+end
+function _rhat(
+    ::Val{:basic}, chains::AbstractArray{<:Union{Missing,Real},3}; split_chains::Int=2
+)
+    # compute size of matrices (each chain may be split!)
+    niter = size(chains, 1) ÷ split_chains
```
devmotion, line 347: It's a bit annoying that the code has (?) to be duplicated. Hopefully at least the tests make it unlikely that they diverge in a noticeable way.

src/ess_rhat.jl
```diff
@@ -239,19 +257,184 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+        return _ess(kind, samples; kwargs...)
+    end
 end
-function ess_rhat(f, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
-    x = _expectand_proxy(f, samples)
+function _ess(estimator, samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
+    x = _expectand_proxy(estimator, samples)
     if x === nothing
-        throw(ArgumentError("the estimator $f is not yet supported by `ess_rhat`"))
+        throw(ArgumentError("the estimator $estimator is not yet supported by `ess`"))
```
devmotion, line 276: Makes me wonder, would it be more natural to add a default fallback for `_expectand_proxy` that does not return `nothing` but throws the descriptive error message directly?

src/mcse.jl
```diff
@@ -2,20 +2,21 @@ const normcdf1 = 0.8413447460685429  # StatsFuns.normcdf(1)
...
+of shape `(draws, chains, parameters)`.
 
-See also: [`ess_rhat`](@ref)
+See also: [`ess`](@ref)
 
-## Estimators
+## Kinds of MCSE estimates
 
-`estimator` must accept a vector of the same `eltype` as `samples` and return a real estimate.
+The estimator whose MCSE should be estimator is specified with `kind`. `kind` must accept a
```
devmotion, line 14:
```suggestion
The estimator whose MCSE should be estimated is specified with `kind`. `kind` must accept a
```

src/ess_rhat.jl
```diff
@@ -239,19 +257,184 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
     doi: [10.1214/20-BA1221](https://doi.org/10.1214/20-BA1221)
     arXiv: [1903.08008](https://arxiv.org/abs/1903.08008)
 """
-function ess_rhat(samples::AbstractArray{<:Union{Missing,Real},3}; kwargs...)
-    return ess_rhat(Statistics.mean, samples; kwargs...)
+function ess(samples::AbstractArray{<:Union{Missing,Real},3}; kind=:bulk, kwargs...)
+    if kind === :bulk
+        return _ess(Val(:bulk), samples; kwargs...)
```
devmotion, line 262: Maybe add a comment to explain why these branches exist and we not just use `Val(kind)`?

src/ess_rhat.jl
```diff
@@ -239,19 +257,184 @@ The ESS and ``\\widehat{R}`` values can be computed for the following estimators
...
+    Rank-normalization, folding, and localization: An improved ``\\widehat {R}`` for
+    assessing convergence of MCMC. Bayesian Analysis.
+    doi: [10.1214/20-BA1221](https://doi.org/10.1214/20-BA1221)
+    arXiv: [1903.08008](https://arxiv.org/abs/1903.08008)
+"""
+function rhat(
+    samples::AbstractArray{<:Union{Missing,Real},3}; kind::Symbol=:rank, kwargs...
+)
+    if kind === :rank
+        return _rhat(Val(:rank), samples; kwargs...)
```
devmotion, line 332: Same here, maybe add a comment?

test/ess_rhat.jl
```diff
@@ -32,7 +32,138 @@ function LogDensityProblems.capabilities(p::CauchyProblem)
     return LogDensityProblems.LogDensityOrder{1}()
 end
 
-@testset "ess.jl" begin
+mymean(x) = mean(x)
+
+@testset "ess_rhat.jl" begin
+    @testset "ess/ess_rhat/rhat basics" begin
+        @testset "only promote eltype when necessary" begin
+            @testset for kind in [:rank, :bulk, :tail, :basic]
```
devmotion, line 40: Maybe use a tuple instead of a vector (also below)?
```suggestion
            @testset for kind in (:rank, :bulk, :tail, :basic)
```

devmotion (commented):
Looks quite good, I only have a few minor comments left.

src/ess_rhat.jl
sethaxen: Since the error includes the function name, it would seem odd to me to put that error in `_expectand_proxy`, and the traceback would then point to this internal function not relevant to the user. Since this function is meant only for internal use, I don't think it's a problem for it to have this fallback.

src/ess_rhat.jl
sethaxen: Yeah, I think we may need a refactor of the code that computes the various variances when we tackle #23, and hopefully then we can eliminate at least part of the redundancy. But yes, the tests should prevent divergences.

src/ess_rhat.jl
devmotion: The second argument is slightly weak though, I think, since also currently the traceback points to an internal function.

sethaxen: True, but `_ess` is more similar to `ess` than  `_expectand_proxy` is.

devmotion (approved):

PR #74, opened 2023-02-26
Make `ess_rhat` return a `NamedTuple`

sethaxen:
This PR changes the return type of `ess_rhat` to a `NamedTuple` from a `Tuple`. This is more consistent with the other diagnostics that return multiple values (e.g. `gelmandiag`), and as a bonus the return value is a valid Table.

### DIFF
--- a/src/ess_rhat.jl
+++ b/src/ess_rhat.jl
@@ -472,7 +472,7 @@ function _ess_rhat(
     ess = similar(chains, T, axes_out)
     rhat = similar(chains, T, axes_out)
 
-    T === Missing && return ess, rhat
+    T === Missing && return (; ess, rhat)
 
     # define caches for mean and variance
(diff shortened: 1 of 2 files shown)


src/ess_rhat.jl
```diff
@@ -565,7 +569,7 @@ function _ess_rhat(
         ess[i] = min(ntotal / τ, ess_max)
     end
 
-    return ess, rhat
+    return (ess=ess, rhat=rhat)
```
devmotion, line 572: A bit simpler:
```suggestion
    return (; ess, rhat)
```

devmotion (approved):

src/ess_rhat.jl
sethaxen: Ah, I had forgotten this was already supported on v1.6.

PR #76, opened 2023-02-27
Fix issues with MLJDecisionTreeInterface

devmotion:
This PR fixes https://github.com/TuringLang/MCMCChains.jl/issues/400.

@ablaom, could you have a quick look at the changes in src/rstar.jl and tell me whether they seem correct to you?

---

https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/76/commits/22ebb26146fac34b276f4ca938a8ad40b696f7bc: Successfully reproduced the problem with MLJDecisionTreeInterface observed in MCMCChains

https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/76/commits/ee1b4f8439e0d55ae5c9a50fa3909d35213f8209: Fix use of MLJ interface (`reformat` and `selectrows`)


src/rstar.jl
```diff
@@ -43,18 +43,21 @@ function rstar(
...
+    fitresult, _ = MMI.fit(classifier, verbosity, train_data...)
 
     # compute predictions on test data
-    xtest, = MMI.selectrows(classifier, test_ids, xdata)
-    ytest = ycategorical[test_ids]
-    predictions = _predict(classifier, fitresult, xtest)
+    # we exploit that MLJ demands that
+    # reformat(model, args...)[1] = reformat(model, args[1])
+    # (https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Implementing-a-data-front-end)
+    test_data = MMI.selectrows(classifier, test_ids, data[1])
```
sethaxen, line 56: Is it guaranteed that the first output of `reformat` corresponds to `x` values?

devmotion: Yep, at least according to the documentation linked above. The more explicit form would be
```suggestion
    data2 = MMI.reformat(classifier, xtable)
    test_data = MMI.selectrows(classifier, test_ids, data2...)
```
but due to the property `reformat(model, args...)[1] = reformat(model, args[1])` we can avoid the second `reformat` call.

ablaom: yes, which is what comment at 54 explains, if I believe.

ablaom (approved):
Looks good to me.

sethaxen (approved):
LGTM

PR #81, opened 2023-04-16
Add option to compute relative ESS

### LINKED ISSUE #80: Add `relative_efficiency`?
It is often useful for other packages (e.g. PSIS, ParetoSmooth) to have a utility function that computes the relative efficiency of draws, effectively `ess(x) ./ sample_size(x)`. We could add this function here and to our API for these downstream packages to use.

Alternatively, like Python ArviZ, we could think of this as a "relative ESS" and instead add a `relative::Bool` keyword to `ess` and `ess_rhat` that, if set to `true`, would return this instead of absolute ESS.

sethaxen:
Fixes #80 by adding a convenience keyword `relative=false` to `ess` and `ess_rhat`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MCMCDiagnosticTools"
 uuid = "be115224-59cd-429b-ad48-344e309966f0"
 authors = ["David Widmann"]
-version = "0.3.1"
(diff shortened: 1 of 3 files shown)


devmotion (approved):
Looks good to me.

PR #86, opened 2023-05-24
DO NOT MERGE - Trigger CI

devmotion:
I saw errors with MLJXGBoostInterface 0.3.8 in some private downstream package. I was wondering if it is possible to reproduce them by running our test suite here.


devmotion:
Seems the downstream issues can't be reproduced by our tests here. Closing for now.

PR #119, opened 2024-02-02
Update ess_hat to not error for short chains

### LINKED ISSUE #117: Make ess_rhat return NaN instead of erroring
Currently MCMCChains automatically displays our diagnostics for `Chains`. But for chains with less than 10 draws, each diagnostic that calls `ess_rhat` will error at the following line due to splitting: https://github.com/TuringLang/MCMCDiagnosticTools.jl/blob/7d78e5d633c4e359be35b39ae2985c52aab16429/src/ess_rhat.jl#L465

This erroring is a nuisance to handle in either PosteriorStats or MCMCChains. I suggest we instead return `NaN`s in such cases and raise an informative warning about why `ess_rhat` failed. This would probably resolve https://github.com/TuringLang/MCMCChains.jl/issues/447 and https://github.com/TuringLang/MCMCChains.jl/issues/410

sethaxen:
Fixes #117. As a side benefit, `ess_rhat` for the `rhat` now returns the same thing as `rhat` even for few draws.

### DIFF
--- a/src/ess_rhat.jl
+++ b/src/ess_rhat.jl
@@ -462,7 +462,7 @@ function _ess_rhat(
     # leave the last even autocorrelation as a bias term that reduces variance for
     # case of antithetical chains, see below
     if !(niter > 4)
-        throw(ArgumentError("number of draws after splitting must >4 but is $niter."))
+        @warn "number of draws after splitting must be >4 but is $niter. ESS cannot be computed."
     end
     maxlag > 0 || throw(DomainError(maxlag, "maxlag must be >0."))
     maxlag = min(maxlag, niter - 4)
@@ -517,6 +517,12 @@ function _ess_rhat(
 
         # estimate rhat
         rhat[i] = sqrt(var₊ / W)
+
+        # only estimate ESS if we have sufficient draws
+        if !(niter > 4)
+            ess[i] = T(NaN)
+            continue
+        end
 
         # center the data around 0
         samples .-= chain_mean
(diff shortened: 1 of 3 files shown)


sethaxen:
The remaining error on the MCMCChains integration test is expected.

src/ess_rhat.jl
```diff
@@ -94,7 +94,7 @@ function build_cache(::FFTAutocovMethod, samples::Matrix, var::Vector)
 
     # create cache for FFT
     T = complex(eltype(samples))
-    n = nextprod([2, 3], 2 * niter - 1)
+    n = nextprod([2, 3], 2 * max(1, niter) - 1)
```
devmotion, line 97: Since we skip computation of ESS in the special case `niter <= 4` anyway, wouldn't it be easier to also skip computing `esscache` and `rel_ess_max`? Possibly it's still nice/an improvement to handle the case `niter < 1` more gracefully here.

sethaxen: I refactored the code to separate allocation of outputs from the allocation of caches with computation of values. When number of iterations is too low, this allows us to directly call the mutating rhat function, which requires no caches.

devmotion (approved):
LGTM if tests pass.

sethaxen:
@devmotion it seems I don't have permissions to merge even with approval because the required integration test doesn't pass (expected). Could my role be upgraded to Admin?

devmotion:
@sethaxen I made you an admin and changed the settings such that admins may bypass the branch protection requirements.

sethaxen:
Thanks!

PR #120, opened 2024-02-13
Update dependency lower bounds and test with CI

sethaxen:
As suggested by https://www.stochasticlifestyle.com/semantic-versioning-semver-is-flawed-and-downgrade-ci-is-required-to-fix-it/, this PR adds a CI step that downgrades all dependencies to their lower bounds before running CI. 

This required bumping some lower bounds that were incompatible with other dependencies. Most of these are within the same version series by semver, the exception being that we drop support for SpecialFunctions `0.8, 0.9, 0.10`. This should be fine though, since the current version is `2.x.x`.


devmotion (commented):
IMO the downgrade CI is a bit restrictive - it's not a problem if Pkg can't resolve all lower bounds at the same time, as long as for each lowest bound the oldest version of packages that it can be resolved with passes the tests, is it?

Nevertheless, I think the checks are helpful. The lower bounds of docs/Project.toml are not tested though, are they (just curious due to the update of docs/Project.toml)?

sethaxen:
> IMO the downgrade CI is a bit restrictive - it's not a problem if Pkg can't resolve all lower bounds at the same time, as long as for each lowest bound the oldest version of packages that it can be resolved with passes the tests, is it?

Agreed, it's just that this would require too many checks. IMO the downgrade approach makes sense if the changes it requires are not too strict. If there's a dependency (e.g. SpecialFunctions) that we really don't want to downgrade, we could of course tell the action to skip it. But we wouldn't be able to test older versions of SpecialFunctions without lowering the lower bounds of Distributions and StatsFuns, and at that point, we wouldn't be able to simultaneously test these lower bounds.

> Nevertheless, I think the checks are helpful. The lower bounds of docs/Project.toml are not tested though, are they (just curious due to the update of docs/Project.toml)?

No, and they would require some updates. The only change I made there was to match the lower bounds used for the package/test, since I think it generally makes sense to keep these consistent.

sethaxen:
Locally everything passes if I downgrade both `Project.toml` and `test/Project.toml`, but the action only downgrades `Project.toml`: https://github.com/cjdoris/julia-downgrade-compat-action/issues/3

I'll troubleshoot.

devmotion:
Regarding test/Project.toml - could we switch back to an extras section in Project.toml? IIRC test/Project.toml was at some point useful to work around issues with test dependencies and the Pkg reaolver but maybe these problems are fixed?

devmotion:
I think I tend to view the doc environment as a "dynamic" environment rather than a package: version bounds are useful to avoid breakages due to a dependency (and CompatHelper notifies us about breaking releases) but we're only interested in building docs with the latest compatible releases of the dependencies and don't want to rebuild the docs with older versions. Hence I think we could generally remove any older semver-breaking versions in the docs, and we could even update it to the latest versions only (but IMO the latter is not neccessary).

sethaxen:
IIUC, that would mean at least bumping the docs versions as suggested in https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/120#issuecomment-1942802430, but leaving `Project.toml` as is. I prefer this also.

devmotion (approved):

PR #128, opened 2024-11-17
Make MCSE for constant arrays always return a NaN

### LINKED ISSUE #122: Making MCSE for constant arrays consistent
Currently `mcse` behaves inconsistently for constant arrays depending on the estimator and method:

```julia
julia> using MCMCDiagnosticTools, Statistics

julia> x = ones(100);

julia> mcse(x; kind=mean)
NaN

julia> mcse(x; kind=std)
NaN

julia> mcse(x; kind=median) # hangs because ESS is NaN, which prevents `betainvcdf` from terminating
^CERROR: InterruptException:
...

julia> mcse(x; kind=x -> mean(x))  # using sbm
0.0

julia> mcse(x; kind=x -> std(x))  # using sbm
0.0

julia> mcse(x; kind=x -> median(x))  # using sbm
0.0
```

We could either:
- return 0.0 for constant arrays
- return NaN for constant arrays

I feel that returning `NaN` makes more sense. Constant arrays are most likely to occur either when a sampled parameter has some fixed entries (e.g. the diagonal of a correlation matrix) or when all transitions that perturbed the parameter were rejected; this often indicates failed sampling. Since MCSE is primarily useful as a convergence diagnostic, I think returning `NaN` makes the most sense here.

sethaxen:
Fixes #122


sethaxen:
#129 should be merged first

src/mcse.jl
```diff
@@ -94,7 +94,7 @@ function _mcse(
 end
 
 function _mcse_quantile(x, p, Seff)
-    Seff === missing && return missing
+    (Seff === missing || isnan(Seff)) && return Seff
```
devmotion, line 97: Given that the other branch returns `(x[upper] - x[lower])/2`, it seems this could easily lead to type instabilities? I think something like

```suggestion
    Seff === missing && return missing
    if isnan(Seff)
        T = eltype(x) / 1
        return T(NaN)
    end
```

would be safer.

src/mcse.jl
```diff
@@ -133,6 +133,10 @@ function _mcse_sbm(f, x, batch_size)
     any(x -> x === missing, x) && return missing
     n = length(x)
     i1 = firstindex(x)
+    if all(==(first(x)), x)
```
devmotion, line 136:
```suggestion
    if allequal(x)
```

sethaxen: Since `allequal` was introduced in Julia v1.8, we would need to either drop support for Julia v1.6-1.7 or add Compat to use it.

devmotion: I think we should use `allequal` at least on newer Julia versions since there's no reason to avoid the specialized implementations for ranges or other special types. I assume if we don't use `allequal` right away, then it's very likely we'll never switch to the better/native approach even when support for older Julia versions is dropped.

I'd be fine with dropping support for 1.6 and 1.7 now that 1.10 is the new LTS (I'd even be fine with dropping support for < 1.10). Alternatively, to avoid adding a Compat dependency, we could also add our own custom, less-optimized function
```julia
if VERSION < v"1.8"
    _allequal(x) = isempty(x) || all(isequal(first(x)), x)
else
    const _allequal = allequal
end
```

sethaxen: I'm okay with bumping support to v1.8. I see no problem with supporting v1.8 and v1.9 until it becomes at all cumbersome to do so.

.github/workflows/CI.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - '1.8'
           - '1'
```
devmotion, line 18: Maybe a good opportunity to switch to
```suggestion
          - 'min'
          - 'lts'
          - '1'
```

.github/workflows/CI.yml
```diff
@@ -14,7 +14,7 @@ jobs:
...
           - 'nightly'
```
devmotion, line 19: Slightly more stable (and relevant?) would be to test pre-releases:
```suggestion
          - 'pre'
```

PR #131, opened 2024-12-02
Add docs interlinks

sethaxen:
This PR adds docs interlinks using DocumenterInterLinks, linking out to relevant Statistics/StatsBase methods and to sections of the MLJ docs.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,4 +1,5 @@
 using Documenter
+using DocumenterInterLinks
 
 # Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
 if haskey(ENV, "GITHUB_ACTIONS")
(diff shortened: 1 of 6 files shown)


sethaxen:
The main downside I see to this approach is that links in the REPL are somewhat less pretty/informative. e.g.

```julia
julia> """
           my_mean(x)
           
       My version of [`Statistics.mean`](@extref). See [Statistics docs](@extref Statistics `Statistics`).

       See also: [`my_std`](@ref).
       """
       function my_mean end
my_mean

help?> my_mean
search: my_mean

  my_mean(x)

  My version of Statistics.mean (@extref). See Statistics docs (@extref Statistics `Statistics`).

  See also: my_std.
```

Still, I think this is better than the alternative of not linking out to other docs.

docs/make.jl
```diff
@@ -11,6 +12,15 @@ DocMeta.setdocmeta!(
     MCMCDiagnosticTools, :DocTestSetup, :(using MCMCDiagnosticTools); recursive=true
 )
 
+links = InterLinks(
+    "MLJ" => "https://juliaai.github.io/MLJ.jl/stable/",
+    "Statistics" => "https://docs.julialang.org/en/v1/",
+    "StatsBase" => (
+        "https://juliastats.org/StatsBase.jl/stable/",
+        "https://juliastats.org/StatsBase.jl/dev/objects.inv",
```
devmotion, line 20: Can this be removed once there is a stable release of StatsBase with docs built with Documenter >= 1?

devmotion (approved):

docs/make.jl
sethaxen: Yes it can. In the meantime, it's harmless, because DocumenterInterLinks will use the first valid inventory, so as soon as the stable one is available, it'll point to that.

PR #147, opened 2025-12-01
Use native Julia binaries in CI

devmotion:
Currently, on MacOS tests are performed with the x64 Julia binary. This can lead to surprising test failures and most likely also doesn't reflect the common user setup.


sethaxen (approved):
workflow failures seem unrelated to this change

PR #150, opened 2025-12-15
chore: Project and workflow maintenance

sethaxen:
This PR makes a number of improvements to the environments and workflows to modernize the package, including:
- Replace `[extras]` with a test-specific environment (`[extras]` will be [deprecated in Julia v1.13](https://pkgdocs.julialang.org/dev/toml-files/#The-[extras]-section-(legacy)), and this is now the recommended approach)
- Add a `test` workspace for Julia >= v1.12
- Add a fallback `[sources]` entry to the test environment for Julia v1.11
- Add a `[sources]` entry to the docs environment
- Replace CompatHelper with DependaBot for managing dependency compats. DependaBot is more convenient when the same dependency compat might need to be bumped for more than one environment. We could keep CompatHelper as well for now, at the cost of more redundant PRs.
- Split the downgrade workflow out from CI to its own workflow so that a failure in this workflow doesn't effect our project CI status. Because of the above changes to the test project, the test and main environments are now jointly resolved by the downgrade action, so unlike before, now we're actually testing at the lower bounds of our test deps. To make this work, Aqua's lower bound needed to be bumped.


devmotion:
Thank you! I just have a few questions:

> Replace [extras] with a test-specific environment ([extras] will be deprecated in Julia v1.13, and this is now the recommended approach)

Does this have any impact on Julia < 1.13?

In my experience separate test environments were often less reliable/stable than `[extras]`, probably due to how Pkg resolved these environments. It seems the use of `[extras]` will be deprecated in 1.13 and users are supposed to use workspaces instead - but those are not supported on Julia 1.10 IIRC.

This makes me wonder whether we should wait with the switch until all supported Julia versions support workspaces?

> Add a test workspace for Julia >= v1.12
> Add a fallback [sources] entry to the test environment for Julia v1.11
> Add a [sources] entry to the docs environment

Also all of these are not supported on Julia 1.10 it seems?

> Replace CompatHelper with DependaBot for managing dependency compats. DependaBot is more convenient when the same dependency compat might need to be bumped for more than one environment. We could keep CompatHelper as well for now, at the cost of more redundant PRs.

Have you used dependabot for Julia compat updates? It's still a beta version, and while I haven't used it myself I saw a few issues reported on the Julia Slack (Guillaume ran into a few it seems, eg https://julialang.slack.com/archives/C6A044SQH/p1765873070848149?thread_ts=1765873070.848149&cid=C6A044SQH and https://julialang.slack.com/archives/C6A044SQH/p1763719297004809?thread_ts=1763719297.004809&cid=C6A044SQH).

That makes me wonder whether we should wait a bit with switching to dependabot until these initial quirks are fixed?

> Split the downgrade workflow out from CI to its own workflow so that a failure in this workflow doesn't effect our project CI status.

That's a very reasonable change IMO.

sethaxen:
> > Replace [extras] with a test-specific environment ([extras] will be deprecated in Julia v1.13, and this is now the recommended approach)
> 
> Does this have any impact on Julia < 1.13?

No, older versions supported the test environment as well. Workspace and sources will be ignored, but the julia test action already executes the tests in a way that works for these older versions.

> In my experience separate test environments were often less reliable/stable than `[extras]`, probably due to how Pkg resolved these environments. It seems the use of `[extras]` will be deprecated in 1.13 and users are supposed to use workspaces instead - but those are not supported on Julia 1.10 IIRC.

I don't recall having issues with stability of test environments. My main complaint was that it was annoying and error-prone to keep test dep compats and project dep compats synchronized, and CompatHelper is not useful here since it doesn't merge PRs. That's why this PR switches to Dependabot, which updates for all projects in a single PR. That has its own downside (see below) but helps with this problem.

> This makes me wonder whether we should wait with the switch until all supported Julia versions support workspaces?

I don't see a reason to. Workspaces are just ignored for older versions, so we don't lose anything by using it. The only possible issue here would be if a user on Julia v1.10 activated `test` and then ran `include("test/runtests.jl")` without first developing `.`. Then they'd run with the latest release instead of the local version of the package, but I don't see this being a serious problem. On the other hand, by switching like this, the downgrade resolver actually downgrades the test-specific dependencies to their lower bounds, which it didn't do beforehand.

> > Add a test workspace for Julia >= v1.12
> > Add a fallback [sources] entry to the test environment for Julia v1.11
> > Add a [sources] entry to the docs environment
> 
> Also all of these are not supported on Julia 1.10 it seems?

Correct, but again, they're just ignored for v1.10, so adding them doesn't break anything.

> Have you used dependabot for Julia compat updates? It's still a beta version, and while I haven't used it myself I saw a few issues reported on the Julia Slack (Guillaume ran into a few it seems, eg https://julialang.slack.com/archives/C6A044SQH/p1765873070848149?thread_ts=1765873070.848149&cid=C6A044SQH and https://julialang.slack.com/archives/C6A044SQH/p1763719297004809?thread_ts=1763719297.004809&cid=C6A044SQH).

Yes I've been using it for a few projects and I'm quite happy with it. In particular, it updates all environments together, so if we have a dep like MLJModelInterface in test and docs, we get a single PR updating it in both. A downside here is that if more than 1 dependency has a new release on a given day, we get a single PR updating both dependencies, but this seems rare. Regarding the issues above, the first seems rare (and perhaps already diagnosed), and the second is already resolved.

> That makes me wonder whether we should wait a bit with switching to dependabot until these initial quirks are fixed?

I think it's polished enough to already be of use. A compromise could be to keep CompatHelper in parallel with dependabot until it's out of beta (I'm doing this for some of my repos alkready).

I do see an issue here I'm going to investigate a bit more, since the resolution process used by the downgrade action seems to ignore the sources and workspace entries. So if we set an MCMCDiagnosticTools compat in `test` that is incompatible with the main project, we should get an error, but instead the main Project manifest actually uses a version of MCMCDiagnosticTools downloaded from General instead of the local version. That makes me concerned that in general it is never using the local version of the package. I'm trying to find the edges of this to open an issue, so for now this PR should remain open.

shravanngoswamii:
@sethaxen Do you mind if I take over this PR?

sethaxen:
> @sethaxen Do you mind if I take over this PR?

I don't mind, but in terms of project features, this PR is already complete. The only reason it hasn't been merged yet is due to the downgrade action having poor support for workspaces/sources. Specifically, this problem persists:

> The downgrad-compat action now has some support for jointly resolving the test env with the main project env, but I uncovered some remaining problems while updating this PR (specifically, the 3rd issue in [julia-actions/julia-downgrade-compat#38](https://github.com/julia-actions/julia-downgrade-compat/issues/38)), so we should wait for that to be resolved before merging.

So there are several ways forward:
- Fix downgrade-compat.
- Split this PR into separate PRs. e.g. switching to dependabot can be done separately, as can splitting the downgrade CI to its own workflow. For the rest, I think you'd need downgrade-compat to be fixed first.
- Accept that downgrade fails for now and merge anyways. Not a big fan of this, because then we miss the signal if a change breaks compatibility with an older dep version.

sethaxen:
All the issues in julia-actions/julia-downgrade-compat#38 have now been resolved, and our downgrade workflow passes. This PR is now ready for review.

.github/dependabot.yml
```diff
@@ -5,3 +5,8 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "monthly"
+  - package-ecosystem: "julia"
+    directories: # Location of Julia projects
+      - "/"
```
devmotion, line 10: Is this sufficient to keep `/`, `/docs` and `/test` updated? Or should the other directories be included here as well?

shravanngoswamii: Nope, they need to be added separately (e.g. https://github.com/TuringLang/Turing.jl/blob/602a69f0fd23b99a22e3000a0a8fbe8e961bb70a/.github/dependabot.yml)

devmotion: No, seems it's not needed - if one uses workspaces as done in this PR: https://discourse.julialang.org/t/psa-github-dependabot-now-supports-julia/134997#p-628604-keeping-test-docs-etc-updated-1

devmotion (approved):

.github/dependabot.yml
shravanngoswamii: Cool

.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1'
+        os:
+          - ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
```
shravanngoswamii:
```suggestion
      - uses: actions/checkout@v7
```

shravanngoswamii: Dependabot will eventually open PR for this

shravanngoswamii (approved):

PR #155, opened 2026-04-07
Change alpha parameter default value to 0.05

penelopeysm:
I think the docstring has a typo here


src/gelmandiag.jl
penelopeysm, line 1: (the default value is here)

devmotion:
I wonder if we should rename `alpha` to `level`, as `level` is arguably a bit clearer and seems to be more commonly used in eg HypothesisTests.

